PR #125, opened 2024-06-22
feat: model worker

jorgeantonio21:
# Introduce ModelWorker for vLLM Model Execution

## Motivation
To improve the efficiency and scalability of our LLM inference service, we're introducing a `ModelWorker` structure. This change aims to better manage model execution, cache handling, and tensor preparation for vLLM inference.

## Description
This PR introduces several key components:

1. `ModelWorker`: A new structure responsible for managing a single LLM model instance (or partition). It handles:
   - Cache management
   - Input tensor preparation
   - Model execution
   - Sampling from model outputs

2. `CacheEngine`: A component within `ModelWorker` that manages KV (key-value) caches for both CPU and GPU. It provides methods for:
   - Allocating cache blocks
   - Swapping blocks between CPU and GPU
   - Copying blocks within GPU memory

3. Enhanced `ModelExecutor` trait: We've expanded this trait to include more detailed model metadata and a default `sample` method implementation.

4. `ModelInput`: A new structure to encapsulate all necessary input data for model forward passes.

5. Utility functions: New helper functions for tensor manipulation and selected token index computation.

6. Error handling: Introduction of specific error types for `ModelWorker` and `CacheEngine`.

7. Tracing: Improved tracing and logging throughout the new components for better observability.

8. Updated dependencies: The PR updates several dependencies, including `candle-core` and `candle-transformers`.

## Breaking Changes
- The `ModelExecutor` trait has been significantly modified. Implementations will need to be updated to conform to the new interface.
- The `LlmService::start` method signature has changed to accommodate the new `ModelWorker` structure.
- The `ExecuteModelRequest` structure now includes additional fields for cache management.

Note: This PR is a substantial change to the model execution pipeline. Thorough testing and gradual rollout are recommended to ensure stability and performance improvements.


atoma-vllm/src/tests/mod.rs
```diff
@@ -222,11 +247,12 @@ async fn test_llm_engine() {
     assert_eq!(number_of_responses, NUM_REQUESTS);
     assert_eq!(elapsed_times.len(), NUM_RUNS);
 
+    // Give enough variability time for different machines
+    let max_elapsed_interval = Duration::from_secs(50);
```
Cifko, line 251: Maybe move these (50, 5 below) to where all the constants are.

atoma-vllm/src/worker.rs
```diff
@@ -0,0 +1,690 @@
...
+
+    /// Swaps cached blocks
+    #[instrument(skip_all)]
+    pub fn cache_swap(
+        &mut self,
+        blocks_to_swap_in: &HashMap<i64, i64>,
+        blocks_to_swap_out: &HashMap<i64, i64>,
+        blocks_to_copy: Option<Tensor>,
+    ) -> Result<(), ModelWorkerError> {
+        if blocks_to_swap_in.len() > 0 {
```
Cifko, line 192: When the memory is getting reorganized it usually means it's almost full, right? That's why we need to swap out some blocks etc. The question is, is it safe to first swap in (as the gpu can be full at this point). Shouldn't we first swap out the blocks from the gpu?

atoma-vllm/src/worker.rs
```diff
@@ -0,0 +1,690 @@
...
+
+                    block_table
+                } else {
+                    // Prefill without chunked prefill
+                    vec![]
+                };
+
+                // 8. Update intermediate states
+                block_tables.push(block_table);
+                sequence_lengths.push(sliding_sequence_length as f32);
```
Cifko, line 323: why are sequence_lengths f32?

atoma-vllm/src/llm_engine.rs
```diff
@@ -29,21 +29,21 @@ use crate::{
 /// in case the `LlmEngine` was on halt.
 const SCHEDULE_WAIT_PERIOD: u64 = 100;
 
-/// `LlmEngine` - An asynchronous worker
-/// to handle scheduling new requests. It is also responsible
+/// `LlmEngine` - An asynchronous worker which is responsible for
+/// scheduling new requests. It is also responsible
 /// to communicate with the `ModelExecutor` service to send new requests
-/// for batched AI inference
+/// for continously batched AI inference
```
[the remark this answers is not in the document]
jorgeantonio21: Great catch !

atoma-vllm/src/scheduler.rs
```diff
@@ -167,9 +167,9 @@ pub struct SchedulerRunningOutputs {
     // Sequences that are swapped out.
     swapped_out: Vec<SequenceGroup>,
     // The blocks to swap out.
-    blocks_to_swap_out: HashMap<u64, u64>,
+    blocks_to_swap_out: HashMap<i64, i64>,
```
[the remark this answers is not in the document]
jorgeantonio21: Actually, this is a bug. The type should be `u32` to be consistent with the respective cuda kernels..

atoma-vllm/src/validation.rs
```diff
@@ -87,6 +87,10 @@ impl Validation {
             encoding.len()
         };
 
+        if input_len == 0 {
+            // TODO: handle the case in which input length == 0
```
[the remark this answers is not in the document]
jorgeantonio21: I am not certain what the behaviour of the system should be in the case of `input_len == 0`.. A model can still generate a response in this case, even though it is likely not what the user intended (this has a cost, right?).. Therefore, it might just be more appropriate for the node to simply commit to an empty request, instead.

atoma-vllm/src/worker.rs
jorgeantonio21: That's super annoying, it is necessary to handle `cumsum` on the defined tensor below.. as it seems that `matmul` is not defined for integer `dtype`s

atoma-vllm/src/worker.rs
jorgeantonio21: Does it really matter ? Both the CPU/GPU memory is allocated at initialization, with a zeroed `KV` cache. This means that whenever we `swap` operations, we are not necessarily allocating new memory, we are just writing from CPU<->GPU. Once the algorithm chooses what has to swap in/out, these operations should be quite independent.. 

What do you think ?

Cifko: Swapping it means moving from cpu to gpu, right? What if all you gpu blocks are used? There is no place to swap them into.

Cifko (approved):

PR #130, opened 2024-08-19
feat: input prompt is store on firebase

Cifko:
Input prompt is now stored on firebase.
This introduces input manager (currently only firebase), that communicates with the atoma-node module.

### DIFF
--- a/atoma-event-subscribe/sui/src/main.rs
+++ b/atoma-event-subscribe/sui/src/main.rs
@@ -20,36 +20,36 @@ struct Args {
 
 #[tokio::main]
 async fn main() -> Result<(), SuiSubscriberError> {
-    tracing_subscriber::fmt::init();
+    //     tracing_subscriber::fmt::init();
 
-    let args = Args::parse();
-    let package_id = ObjectID::from_hex_literal(&args.package_id)?;
-    let http_url = args.http_addr;
-    let ws_url = args.ws_addr;
+    //     let args = Args::parse();
+    //     let package_id = ObjectID::from_hex_literal(&args.package_id)?;
+    //     let http_url = args.http_addr;
+    //     let ws_url = args.ws_addr;
 
-    let (event_sender, mut event_receiver) = tokio::sync::mpsc::channel(32);
+    //     let (event_sender, mut event_receiver) = tokio::sync::mpsc::channel(32);
 
-    let event_subscriber = SuiSubscriber::new(
-        1,
-        &http_url,
-        Some(&ws_url),
-        package_id,
-        event_sender,
-        Some(Duration::from_secs(5 * 60)),
-    )
-    .await?;
+    //     let event_subscriber = SuiSubscriber::new(
+    //         1,
+    //         &http_url,
+    //         Some(&ws_url),
+    //         package_id,
+    //         event_sender,
+    //         Some(Duration::from_secs(5 * 60)),
+    //     )
+    //     .await?;
 
-    tokio::spawn(async move {
-        info!("initializing subscribe");
-        if let Err(err) = event_subscriber.subscribe().await {
-            error!("Failed to subscribe: {:?}", err);
-        }
-        Ok::<_, SuiSubscriberError>(())
-    });
+    //     tokio::spawn(async move {
+    //         info!("initializing subscribe");
+    //         if let Err(err) = event_subscriber.subscribe().await {
+    //             error!("Failed to subscribe: {:?}", err);
+    //         }
+    //         Ok::<_, SuiSubscriberError>(())
+    //     });
 
-    while let Some(event) = event_receiver.recv().await {
-        info!("Processed a new event: {:?}", event);
-    }
+    //     while let Some(event) = event_receiver.recv().await {
+    //         info!("Processed a new event: {:?}", event);
+    //     }
 
     Ok(())
 }
--- a/atoma-event-subscribe/sui/src/main.rs
+++ b/atoma-event-subscribe/sui/src/main.rs
@@ -20,36 +20,45 @@ struct Args {
 
 #[tokio::main]
 async fn main() -> Result<(), SuiSubscriberError> {
-    //     tracing_subscriber::fmt::init();
+    tracing_subscriber::fmt::init();
 
-    //     let args = Args::parse();
-    //     let package_id = ObjectID::from_hex_literal(&args.package_id)?;
-    //     let http_url = args.http_addr;
-    //     let ws_url = args.ws_addr;
+    let args = Args::parse();
+    let package_id = ObjectID::from_hex_literal(&args.package_id)?;
+    let http_url = args.http_addr;
+    let ws_url = args.ws_addr;
 
-    //     let (event_sender, mut event_receiver) = tokio::sync::mpsc::channel(32);
+    let (event_sender, mut event_receiver) = tokio::sync::mpsc::channel(32);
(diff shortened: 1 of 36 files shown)


atoma-event-subscribe/sui/src/main.rs
```diff
@@ -20,36 +20,36 @@ struct Args {
...
-        if let Err(err) = event_subscriber.subscribe().await {
-            error!("Failed to subscribe: {:?}", err);
-        }
-        Ok::<_, SuiSubscriberError>(())
-    });
-
-    while let Some(event) = event_receiver.recv().await {
-        info!("Processed a new event: {:?}", event);
-    }
+    //     tracing_subscriber::fmt::init();
```
jorgeantonio21, line 23: Why ?

atoma-event-subscribe/sui/src/subscriber.rs
```diff
@@ -213,7 +220,24 @@ impl SuiSubscriber {
     #[instrument(skip(self, event_data))]
     async fn handle_prompt_event(&self, event_data: Value) -> Result<(), SuiSubscriberError> {
         debug!("event data: {}", event_data);
-        let request = Request::try_from((self.id, event_data))?;
+        let mut request = Request::try_from((self.id, event_data))?;
+        let metadata = AtomaInputMetadata {
+            user_id: request.params().prompt(),
```
jorgeantonio21, line 225: I don't understand how the `user_id` is the `prompt`, at most it should be its hash..

atoma-event-subscribe/sui/src/subscriber.rs
```diff
@@ -213,7 +220,24 @@ impl SuiSubscriber {
...
+            node_id: self.id,
+            input_source: atoma_types::InputSource::Firebase,
+            input_type: atoma_types::InputType::Text,
+        };
+        let (oneshot_sender, oneshot_receiver) = tokio::sync::oneshot::channel();
+        self.input_manager_tx
+            .send((metadata, oneshot_sender))
+            .await
+            .map_err(Box::new)?;
+        let result = tokio::time::timeout(Duration::from_secs(5), oneshot_receiver)
```
jorgeantonio21, line 236: Let's hardcode the duration in a `const`. Also, isn't this introducing too much latency ? I wonder if we would not be good with a few ms ?

atoma-input-manager/src/firebase/mod.rs
```diff
@@ -0,0 +1,79 @@
...
+                .path_segments_mut()
+                .map_err(|_| AtomaInputManagerError::UrlError("URL is not valid".to_string()))?;
+            path_segment.push("prompts");
+            path_segment.push(&input_metadata.user_id);
+            path_segment.push(&input_metadata.node_id.to_string());
+            path_segment.push(&input_metadata.ticket_id);
+            path_segment.push("prompt.json");
+        }
+        url.set_query(Some(&format!("auth={token}")));
+        dbg!(&url);
```
jorgeantonio21, line 62: Should it be removed ?

atoma-output-manager/src/lib.rs
```diff
@@ -90,8 +91,8 @@ pub enum AtomaOutputManagerError {
     RequestError(#[from] reqwest::Error),
     #[error("GraphQl error: `{0}`")]
     GraphQlError(String),
-    #[error("Invalid output destiny: `{0}`")]
-    InvalidOutputDestiny(String),
+    #[error("Invalid output destination: `{0}`")]
```
jorgeantonio21, line 94: Nice !

atoma-input-manager/src/lib.rs
```diff
@@ -0,0 +1,99 @@
...
+
+/// `AtomaInputManager` - manages different input sources
+///     requests, allowing for a flexible interaction between
+///     the user and the Atoma Network.
+///
+/// Current available options for storing the input on behalf
+/// of the user consists of Firebase.
+pub struct AtomaInputManager {
+    /// Firebase's input manager instance.
+    firebase_input_manager: FirebaseInputManager,
```
jorgeantonio21, line 22: The goal is to add support, over time, for different sources to interact with Atoma, right ?

atoma-input-manager/src/firebase/mod.rs
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    /// Handles  a new post request. Encapsulates the logic necessary
+    /// to post new requests, using `reqwest::Client`.
+    #[instrument(skip_all)]
+    pub async fn handle_get_request(
+        &mut self,
+        input_metadata: AtomaInputMetadata,
+    ) -> Result<String, AtomaInputManagerError> {
+        dbg!(&input_metadata);
```
jorgeantonio21, line 45: Shall it be removed ? Or maybe just add a `debug!` instead

atoma-input-manager/src/firebase/mod.rs
```diff
@@ -0,0 +1,79 @@
...
+        let client = Client::new();
+        let token = self.auth.get_id_token().await?;
+        dbg!(&token);
```
jorgeantonio21, line 48: See previous comment

atoma-input-manager/src/firebase/mod.rs
```diff
@@ -0,0 +1,79 @@
...
+        let mut url = self.firebase_url.clone();
+        dbg!(&url);
```
jorgeantonio21, line 50: See previous comment

atoma-types/src/lib.rs
```diff
@@ -682,10 +722,27 @@ pub struct AtomaOutputMetadata {
     pub leaf_hash: Vec<u8>,
     /// The transaction in base58 format (used in the Sui blockchain)
     pub transaction_base_58: String,
-    /// The output destiny
+    /// The output destination
```
jorgeantonio21, line 725: Nice !

jorgeantonio21 (commented):
Looks great ! 

Left a few comments to be addressed, other than that lgtm

atoma-event-subscribe/sui/src/subscriber.rs
Cifko: because now the event subscriber expects channel from input manager. Why do we even have standalone binaries for this? We will never run just event subscriber.

atoma-event-subscribe/sui/src/subscriber.rs
Cifko: we can finetune this with testing, but I would prefer for the alpha version more timeout instead of failing.

atoma-event-subscribe/sui/src/main.rs
Cifko: because now the event subscriber expects channel from input manager. Why do we even have standalone binaries for this? We will never run just event subscriber.

atoma-event-subscribe/sui/src/subscriber.rs
Cifko: because I did not modify the contract and I had to put the user id somewhere for the time being. And the prompt is now empty.
I renamed the function getter, but the field is still prompt, because it's json serialized.

atoma-input-manager/src/lib.rs
Cifko: Yes.

atoma-event-subscribe/sui/src/main.rs
jorgeantonio21: It is helpful to have internal testing, but we can remove it

PR #142, opened 2024-09-04
fix: no crash on input error

Cifko:
Do not crash on input error (e.g. timeout)
Solves #140

### DIFF
--- a/atoma-event-subscribe/sui/src/subscriber.rs
+++ b/atoma-event-subscribe/sui/src/subscriber.rs
@@ -1,5 +1,6 @@
 use std::{path::Path, str::FromStr, time::Duration};
 
+use atoma_input_manager::AtomaInputManagerError;
 use futures::StreamExt;
 use serde_json::Value;
 use sui_sdk::rpc_types::{EventFilter, SuiEvent};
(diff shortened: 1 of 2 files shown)


atoma-event-subscribe/sui/src/subscriber.rs
```diff
@@ -228,12 +238,7 @@ impl SuiSubscriber {
             .send((request.params().prompt(), oneshot_sender))
             .await
             .map_err(Box::new)?;
-        let result = tokio::time::timeout(
-            Duration::from_secs(WAIT_FOR_INPUT_MANAGER_RESPONSE_SECS),
-            oneshot_receiver,
-        )
-        .await
-        .map_err(|_| SuiSubscriberError::TimeoutError)??;
+        let result = oneshot_receiver.await??;
```
jorgeantonio21, line 241: Why is this better ?

jorgeantonio21 (approved):
LGTM, left a comment though

atoma-event-subscribe/sui/src/subscriber.rs
Cifko: because when the `AtomaInputManager` timeouts it will send the timeout error. So I don't have to have timeout here. Before it just didn't send anything so I had to have the timeout here as well.

jorgeantonio21: But doesn't this way also panics the node service ? You are still crashing on error, no ?

Cifko: No, the error is handled properly on line 201

jorgeantonio21: Agree !

PR #147, opened 2024-09-09
20240903 ipfs input manager

jorgeantonio21:


atoma-output-manager/src/lib.rs
```diff
@@ -70,7 +70,7 @@ impl AtomaOutputManager {
             }
             Err(e) => {
                 error!(
-                    "Failed to obtain IPFS client's version: {e}, most likely IPFS daemon is not running in the background. To start it, run `$ ipfs daemon`"
+                    "Failed to obtain IPFS client's version: {e}, most likely IPFS daemon is not running in the background. To start it, install IPFS daemon on your machine and run `$ ipfs daemon` on Linux, or `ipfs daemon` on Windows."
```
Cifko, line 73: Print the error depending on the system running
something like
```
    #[cfg(target_os = "windows")]
    {
        println!("This is Windows.");
    }

    #[cfg(target_os = "linux")]
    {
        println!("This is Linux.");
    }
```

Cifko (approved):

atoma-output-manager/src/lib.rs
jorgeantonio21: Nice !

PR #152, opened 2024-09-13
feat: add image to firebase

Cifko:
Add image to firebase when generated via text2image models.


atoma-inference/src/models/candle/stable_diffusion.rs
```diff
@@ -256,6 +251,7 @@ impl ModelTrait for StableDiffusion {
             load_data.use_flash_attention,
             load_data.dtype,
         )?;
+        info!("Loaded.");
```
jorgeantonio21, line 254: Maybe add more logging info here ?

jorgeantonio21 (approved):
LGTM

PR #153, opened 2024-09-17
feat: store ipfs cid into firebase alongside with the image

Cifko:
If the output destination is ipfs, store the image in firebase as well, and store the ipfs cid in firebase.


jorgeantonio21 (approved):
LGTM

PR #155, opened 2024-09-20
fix: properly refresh expired token

Cifko:
When the firebase token is about to expire call the proper refresh.
In rare case when the refresh token expires, do a new signup and store the our `small_id` under this new `uid`


jorgeantonio21 (approved):
Great ! 

LGTM

jorgeantonio21:
Add to resolve a few conflicts with the `main` branch before merging, after we merged other PRs first.

PR #162, opened 2024-10-17
feat: atoma state

jorgeantonio21:


atoma-state/src/state_manager.rs
```diff
@@ -0,0 +1,1320 @@
...
+        level = "trace",
+        skip_all,
+        fields(stack_small_id = %stack_small_id)
+    )]
+    pub async fn get_stack_settlement_ticket(
+        &self,
+        stack_small_id: i64,
+    ) -> Result<StackSettlementTicket> {
+        let stack_try_settle =
+            sqlx::query("SELECT * FROM stack_settlement_tickets WHERE stack_small_id = ?")
```
Cifko, line 710: I don't see the `stack_settlement_tickets` table created.

atoma-state/src/state_manager.rs
```diff
@@ -0,0 +1,1320 @@
...
+    ///     state_manager.remove_task(task_small_id).await
+    /// }
+    /// ```
+    #[tracing::instrument(
+        level = "trace",
+        skip_all,
+        fields(task_small_id = %task_small_id)
+    )]
+    pub async fn remove_task(&self, task_small_id: i64) -> Result<()> {
+        sqlx::query("DELETE FROM tasks WHERE task_small_id = ?")
```
Cifko, line 212: does sqlx cascade deletion? because these is foreign key from another table. That question applies to all deletions.

atoma-state/src/state_manager.rs
Cifko, line 1: maybe add some tests.

atoma-state/src/state_manager.rs
jorgeantonio21: That's a big one, thanks for noticing !

atoma-state/src/state_manager.rs
jorgeantonio21: Added

atoma-state/src/state_manager.rs
jorgeantonio21: I removed the removals for now. I think it is good for node operators to have access to their full transactional history. It is also possible to remove manually these tables if they grow up and everything is settled.

Cifko (approved):

PR #169, opened 2024-10-23
feat: add sui client for node transactions

jorgeantonio21:


atoma-sui/src/client.rs
```diff
@@ -0,0 +1,1023 @@
...
+                    Some(SuiObjectResponseQuery {
+                        filter: Some(SuiObjectDataFilter::Package(package)),
+                        options: Some(SuiObjectDataOptions {
+                            show_type: true,
+                            show_content: true,
+                            ..Default::default()
+                        }),
+                    }),
+                    cursor,
+                    Some(100),
```
Cifko, line 976: can we move this constant, to some constant at the top of the file, like `PAGE_SIZE` or something like that.

Cifko (approved):

atoma-sui/src/client.rs
```diff
@@ -0,0 +1,1023 @@
...
+    ///     1,
+    ///     Some(node_badge),
+    ///     Some(gas_object),
+    ///     Some(10_000_000),  // 0.01 SUI
+    ///     Some(1000)
+    /// ).await?;
+    /// # Ok(())
+    /// # }
+    /// ```
+    #[instrument(level = "info", skip_all, fields(
```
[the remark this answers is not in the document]
jorgeantonio21: This makes sense, added a few parameters for logging.

atoma-sui/src/client.rs
jorgeantonio21: Done

PR #187, opened 2024-10-28
feat: change sqlite to any

Cifko:
Change `sqlite` to `any` so it can be reused in the `atoma-proxy` with postgres.


Cifko:
apparently the any driver is not supporting sqlite bool value

PR #195, opened 2024-10-29
fix: rename the db indexes to conventional naming

Cifko:
Rename the indexes to conventional naming to prevent collisions.


jorgeantonio21 (approved):
LGTM

PR #196, opened 2024-10-29
fix: get tokenizer.json

Cifko:
Fix getting tokenizer.json. Use hugging face api to get protected tokenizers.


atoma-bin/atoma_node.rs
```diff
@@ -101,31 +102,34 @@ impl Config {
...
-                    anyhow::anyhow!(format!(
-                        "Failed to parse tokenizer for model {}, with error: {}",
-                        model, e
-                    ))
-                })
-                .map(Arc::new)
+        .map(|(model, revision)| {
+            let api = api.clone();
+            async move {
+                let repo = api.repo(Repo::with_revision(
```
jorgeantonio21, line 114: should we add an optional cache directory here ?

atoma-service/src/config.rs
```diff
@@ -43,6 +43,11 @@ pub struct AtomaServiceConfig {
     ///
     /// This field specifies the address and port on which the Atoma Service will bind.
     pub service_bind_address: String,
+
+    /// Token for the Hugging Face API.
+    ///
+    /// This field specifies the token used to authenticate with the Hugging Face API. That is used to fetch the models tokenizers.
+    pub hf_token: String,
```
jorgeantonio21, line 50: This can come from the .env file (which already is defined there), we don't need to add it to the config.toml

config.example.toml
```diff
@@ -3,6 +3,7 @@ inference_service_url = "http://localhost:8000"
 models = ["meta-llama/Llama-3.1-70b-Instruct"]
 revisions = ["main"]
 service_bind_address = "http://0.0.0.0:8080"
+hf_token = "you hugging face api token"
```
jorgeantonio21, line 6: I think we should not add it here

jorgeantonio21 (approved):
LGTM

I think we should not include the hf token in the config.toml, instead use it from the .env

PR #198, opened 2024-10-29
feat: embeddings inference server

### LINKED ISSUE #182: feat: embeddings inference

francis2tm:
Fixes #182 and #194

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -1,6 +1,7 @@
 services:
   vllm:
-    image: vllm/vllm-openai:latest
+    profiles: [chat_completions_vllm]
+    image: vllm/vllm-openai:v0.6.3
     runtime: nvidia
     deploy:
       resources:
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -75,9 +79,10 @@ ATOMA_SERVICE_PORT=3000       # External port for Atoma service
 ```
 
 3. Configure `config.toml`, using `config.example.toml` as template:
+
 ```toml
 [atoma-service]
-inference_service_url = "http://vllm:8000"    # Internal Docker network URL
+inference_service_url = "http://chat-completions:8000"    # Internal Docker network URL
 embeddings_service_url = ""
```
jorgeantonio21, line 86: Can we add a placeholder value ? Similar to the chat-completions one ?

README.md
```diff
@@ -75,9 +79,10 @@ ATOMA_SERVICE_PORT=3000       # External port for Atoma service
 ```
 
 3. Configure `config.toml`, using `config.example.toml` as template:
+
 ```toml
 [atoma-service]
-inference_service_url = "http://vllm:8000"    # Internal Docker network URL
+inference_service_url = "http://chat-completions:8000"    # Internal Docker network URL
```
jorgeantonio21, line 85: I suggest changing the name of the variable to `chat_completions_service_url`.

docker-compose.yaml
```diff
@@ -58,6 +35,47 @@ services:
     networks:
       - atoma-network
 
+  vllm:
+    container_name: chat-completions
+    profiles: [chat_completions_vllm]
+    image: vllm/vllm-openai:v0.6.3
```
jorgeantonio21, line 41: why not `latest` ? Manually having to track each version is not ideal, and I believe that any potential changes to openai api on vLLM will be prior advertised

start.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
```
jorgeantonio21, line 1: Please, add proper instructions, on the README.md, on how to use the `start.sh` executable at launch, and how to set the right variables, so that each 'profiled' service can start.

jorgeantonio21 (commented):
LGTM, 

Let's address the few comments I left, and we will merge it

docker-compose.yaml
francis2tm: I thought about that, and it's another possible way. But see the following reasons against using `latest`:
- https://vsupalov.com/docker-latest-tag/

For me, I like having the peace of mind to know that **atoma-node won't break from one day to another with the same git commit**. Which is not the case if we rely on `latest` (a bit the same philosophy to using `Cargo.lock`, if we specify the version to of the images, we know for a fact that everyone that runs `atoma-node` on the same git commit, they will have the same build running).

start.sh
francis2tm: @jorgeantonio21 Btw, I'm not very happy with the `start.sh`. 
DO you have a suggestion on how to get rid of it? Many node operators will not like running a bash script (and rightly so). But, at the same time, I didn't see another way to add some logic (e.g. make sure user does not select 2 chat completions backends at the same time. This will be relevant when we have more than 1 backend for chat completions)

docker-compose.yaml
jorgeantonio21: Sounds good, I am happy with it

start.sh
francis2tm: So, I figured out the most optimal way:
- the validation can happen implicitly on the `container_name`. Just tested, and there cannot be 2 containers with the same name, so if user tries to spin 2 containers for chat completions, docker compose will throw that error.
- No bash required

THis is solved in https://github.com/atoma-network/atoma-node/pull/205

francis2tm:
Closing. This this PR's functionality will be added in #205.

PR #199, opened 2024-10-30
feat: add new_from_config to AtomaSuiClient

Cifko:
Add `new_from_config` to `AtomaSuiClient` and change the new to take config as a parameter. So it's the same as other structs that we have in this repo.


jorgeantonio21 (approved):
LGTM

Run `cargo fmt` before we merge it

PR #200, opened 2024-10-30
feat: add sui event cursor saving, to ease the process of rerun nodes  

jorgeantonio21:


jorgeantonio21:
Addresses #173

atoma-sui/src/subscriber.rs
```diff
@@ -261,6 +261,84 @@ impl SuiEventSubscriber {
...
+/// ```
+fn read_cursor_from_toml_file(path: &str) -> Result<Option<EventID>> {
+    match std::fs::read_to_string(path) {
+        Ok(content) => {
+            let cursor: EventID = toml::from_str(&content)?;
+            Ok(Some(cursor))
+        }
+        Err(e) if e.kind() == std::io::ErrorKind::NotFound => {
+            // Create the file if it doesn't exist
+            std::fs::write(path, "")?;
```
Cifko: if you don't plan to write anything into the file, then use `std::fs::File::create_new(path)`

Cifko (approved):

atoma-sui/src/subscriber.rs
jorgeantonio21: I removed this, as reading from empty file doesn't require creating the file, instead we should write the file for the first on shutdown (I reckon). What are your thoughts ?

Cifko: on shutdown? shouldn't we create the file only when needed?

Cifko (approved):

PR #207, opened 2024-10-31
fix: remove page size for cursor

Cifko:
Remove the page size. If `None` is provided, it will use the max available from the sdk. Which currently is 50. So I'm getting : `RpcError(Call(Custom(ErrorObject { code: InvalidParams, message: "Page size limit 100 exceeds max limit 50", data: None })))`
So lets just let the sdk use the max.


jorgeantonio21 (approved):
LGTM

PR #217, opened 2024-11-03
feat: inference service endpoints

### LINKED ISSUE #216: feat: inference server endpoints
Axum endpoints on `atoma-service` that interface with the local inference services

francis2tm:
Fixes #216


atoma-service/src/components/openapi.rs
```diff
@@ -0,0 +1,67 @@
...
+            (name = "embeddings", description = "Embeddings"),
+            (name = "images", description = "Image generations"),
+        ),
+        servers(
+            (url = "http://localhost:8080"),
+        )
+    )]
+    struct ApiDoc;
+
+    // struct SecurityAddon;
```
jorgeantonio21, line 32: Shall we remove this ?

atoma-service/src/middleware.rs
```diff
@@ -251,61 +288,13 @@ pub async fn verify_stack_permissions(
             StatusCode::BAD_REQUEST
         })?;
     if !state.models.contains(&model.to_string()) {
+        info!("Supported models: {:?}", state.models);
```
jorgeantonio21, line 291: It is better to add this to the error below

atoma-service/src/server.rs
```diff
@@ -208,140 +205,24 @@ pub async fn run_server(
...
 /// ```
+
+pub const HEALTH_PATH: &str = "/health";
+#[derive(OpenApi)]
+#[openapi(paths(health))]
+pub(crate) struct HealthOpenApi;
+
 #[utoipa::path(
     get,
-    path = "/health",
```
jorgeantonio21, line 213: why ?

jorgeantonio21 (commented):
LGTM

Address the comments above and we can merge it

atoma-service/src/server.rs
francis2tm: Because the path is already being defined here https://github.com/atoma-network/atoma-node/blob/98d20ddefb9308b4b6694696c98dbbf1f1ab26e6/atoma-service/src/server.rs#L211

PR #226, opened 2024-11-07
fix: sui address instead of pub key

Cifko:
`owner_address` in stacks is type of sui address, not sui public key. This PR gets the sui address from the sui public key used to sign the request.


jorgeantonio21 (approved):
LGTM

PR #229, opened 2024-11-11
feat: add the option to override config from env vars

Cifko:
Add the possibility to override config values from env vars.
E.g.  in the config
```
[atoma_daemon]
service_bind_address
```
can be overwritten with setting `ATOMA_DAEMON__SERVICE_BIND_ADDRESS` (mind the double _ between ATOMA_DAEMON and the rest.
After merging this, change your local config.toml. The section names are changing from dash to underscore, e.g. `atoma-daemon` to `atoma_daemon`


jorgeantonio21 (approved):
LGTM

PR #239, opened 2024-11-15
fix: proper shutdown/failure

Cifko:
now there is global shutdown
- it's triggered when one of the task fails, or ctrl_c is pressed
- it will properly shutdown other tasks

### DIFF
--- a/atoma-bin/atoma_daemon.rs
+++ b/atoma-bin/atoma_daemon.rs
@@ -1,6 +1,6 @@
 use std::{path::Path, str::FromStr, sync::Arc};
 
-use anyhow::Result;
+use anyhow::{Context, Result};
 use atoma_daemon::{
     config::AtomaDaemonConfig,
     daemon::{run_daemon, DaemonState},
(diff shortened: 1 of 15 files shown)


atoma-bin/atoma_daemon.rs
```diff
@@ -33,6 +37,23 @@ struct DaemonArgs {
...
+    shutdown_sender: watch::Sender<bool>,
+) -> tokio::task::JoinHandle<Result<()>>
+where
+    F: std::future::Future<Output = Result<()>> + Send + 'static,
+{
+    tokio::task::spawn(async move {
+        let res = f.await;
+        if res.is_err() {
+            // Only send shutdown signal if the task failed
+            shutdown_sender.send(true).unwrap();
```
jorgeantonio21, line 51: Can we instead handle error here ?

atoma-bin/atoma_daemon.rs
```diff
@@ -67,14 +88,32 @@ async fn main() -> Result<()> {
...
         target = "atoma_daemon",
         event = "atoma-daemon-stop",
         "Atoma daemon service stopped gracefully..."
     );
 
-    Ok(())
+    let ctrl_c = tokio::task::spawn(async move {
+        tokio::select! {
+            _ = tokio::signal::ctrl_c() => {
+                println!("ctrl-c received, sending shutdown signal");
```
jorgeantonio21, line 106: Please replace with `tracing::info!`

atoma-bin/atoma_node.rs
```diff
@@ -153,6 +153,24 @@ async fn initialize_tokenizers(
...
+) -> tokio::task::JoinHandle<Result<()>>
+where
+    E: Into<anyhow::Error>,
+    F: std::future::Future<Output = Result<(), E>> + Send + 'static,
+{
+    tokio::task::spawn(async move {
+        let res = f.await;
+        if res.is_err() {
+            // Only send shutdown signal if the task failed
+            shutdown_sender.send(true).unwrap();
```
jorgeantonio21, line 168: Can we please handle error here ?

atoma-bin/atoma_node.rs
```diff
@@ -153,6 +153,24 @@ async fn initialize_tokenizers(
     try_join_all(fetch_futures).await
 }
 
+fn spawn_with_shutdown<F, E>(
```
jorgeantonio21, line 156: 1. Add docs.
2. This method is repeated, we could add a utils crate for this sort of functionality, but it can be done on a posterior iteration

atoma-bin/atoma_node.rs
Cifko: why should we handle it here, it's handled elsewhere, this just propagates the error

jorgeantonio21 (approved):

PR #245, opened 2024-11-21
feat: migrate from sqlite to postgres

jorgeantonio21:
1. Added proper migrations code;
2. Docker service for postgres;
3. Proper GitHub workflows for testing;
4. Refactor sqlite logic to postgres.


francis2tm (approved):

atoma-service/src/tests.rs
```diff
@@ -83,29 +84,36 @@ mod middleware {
         Tokenizer::from_str(&tokenizer_json).unwrap()
     }
 
+    async fn truncate_tables() {
+        let db = PgPool::connect("postgres://atoma:atoma@localhost:5432/atoma")
```
Cifko, line 88: can the DB used for test, have `test` in the name e.g. `atoma_test` 
also add this url to constant, it's used in two places, it would be easier to rewrite if it's just one place
I know that for the docker test it doesn't matter, but if I want to run them locally with my own postgres instance then it matters.
or create random DB name on each test, and we can run them in parallel

Cifko (approved):

atoma-service/src/tests.rs
jorgeantonio21: Great suggestion ! I went with the single constant assignment, as I am not sure how feasible is to deploy postgres instances for each new url

Issue #228, opened 2024-11-08
When a node receives a new request for a stack, if the node did not have the stack stored in the database, then it should query the blockchain instead


jorgeantonio21:
In the `middleware.rs` file, on the `atoma-service` crate, we query the stack to which the request is charged on, in the method `verify_stack_permissions`:

```rs
let (result_sender, result_receiver) = oneshot::channel();
    state
        .state_manager_sender
        .send(
            AtomaAtomaStateManagerEvent::GetAvailableStackWithComputeUnits {
                stack_small_id,
                sui_address: sui_address.to_string(),
                total_num_tokens,
                result_sender,
            },
        )
        .map_err(|err| {
            error!("Failed to get available stacks: {}", err);
            StatusCode::UNAUTHORIZED
        })?;
    let available_stack = result_receiver
        .await
        .map_err(|_| {
            error!("Failed to get available stack with enough compute units");
            StatusCode::UNAUTHORIZED
        })?
        .map_err(|err| {
            error!(
                "Failed to get available stack with enough compute units: {}",
                err
            );
            StatusCode::UNAUTHORIZED
        })?;
    if available_stack.is_none() {
        error!("No available stack with enough compute units");
        return Err(StatusCode::UNAUTHORIZED);
    }
```

It might be the case that the current node Sui event subscriber hasn't yet catched the buying stack event. In that case, there are two options:

1. Either the middleware waits for a period time (say 300ms), and retries;
2. Or the middleware queries the event subscriber/state manager to read the new events of the blockchain and check directly if the event is available or not.

jorgeantonio21:
Addressed in #248

PR #248, opened 2024-11-23
feat: add blockchain query for newly created stacks that are requested to the atoma service

jorgeantonio21:
Addresses #228

### DIFF
--- a/atoma-service/src/middleware.rs
+++ b/atoma-service/src/middleware.rs
@@ -23,6 +23,7 @@ use serde_json::Value;
 use sui_sdk::types::{
     base_types::SuiAddress,
     crypto::{PublicKey, Signature, SuiSignature},
+    digests::TransactionDigest,
 };
 use tokio::sync::oneshot;
 use tracing::{error, instrument};
@@ -334,8 +335,24 @@ pub async fn verify_stack_permissions(
             StatusCode::UNAUTHORIZED
         })?;
     if available_stack.is_none() {
-        error!("No available stack with enough compute units");
-        return Err(StatusCode::UNAUTHORIZED);
+        let tx_digest_str = req_parts
+            .headers
+            .get("X-Tx-Digest")
+            .ok_or_else(|| {
+                error!("Tx digest header not found");
+                StatusCode::BAD_REQUEST
+            })?
+            .to_str()
+            .map_err(|_| {
+                error!("Tx digest cannot be converted to a string");
+                StatusCode::BAD_REQUEST
+            })?;
+        let tx_digest = TransactionDigest::from_str(tx_digest_str).unwrap();
+        let compute_units = utils::request_blockchain_for_stack(&state, tx_digest).await?;
+        if compute_units > total_num_compute_units as u64 {
+            error!("No available stack with enough compute units");
+            return Err(StatusCode::UNAUTHORIZED);
+        }
     }
     let request_metadata = RequestMetadata::default()
         .with_stack_info(stack_small_id, total_num_compute_units)
@@ -353,7 +370,7 @@ pub(crate) mod utils {
         secp256r1::{Secp256r1PublicKey, Secp256r1Signature},
         traits::{ToFromBytes, VerifyingKey},
     };
-    use sui_sdk::types::crypto::SignatureScheme;
+    use sui_sdk::types::{crypto::SignatureScheme, digests::TransactionDigest};
 
     /// Verifies the authenticity of a request by checking its signature against the provided hash.
     ///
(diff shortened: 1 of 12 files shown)


atoma-service/src/middleware.rs
```diff
@@ -334,8 +335,24 @@ pub async fn verify_stack_permissions(
             StatusCode::UNAUTHORIZED
         })?;
     if available_stack.is_none() {
-        error!("No available stack with enough compute units");
-        return Err(StatusCode::UNAUTHORIZED);
+        let tx_digest_str = req_parts
+            .headers
+            .get("X-Tx-Digest")
+            .ok_or_else(|| {
+                error!("Tx digest header not found");
```
Cifko, line 342: Maybe change this to `Stack not found, tx digest header expected but not found` or something like that, because to someone unfamiliar with the code this error can sound like that tx digest should always be there.

atoma-service/src/middleware.rs
```diff
@@ -334,8 +335,24 @@ pub async fn verify_stack_permissions(
...
+                StatusCode::BAD_REQUEST
+            })?
+            .to_str()
+            .map_err(|_| {
+                error!("Tx digest cannot be converted to a string");
+                StatusCode::BAD_REQUEST
+            })?;
+        let tx_digest = TransactionDigest::from_str(tx_digest_str).unwrap();
+        let compute_units = utils::request_blockchain_for_stack(&state, tx_digest).await?;
+        if compute_units > total_num_compute_units as u64 {
```
Cifko, line 352: you have to verify that the `stack_small_id` in the request matches the `stack_small_id` in the tx digest.

atoma-service/src/middleware.rs
```diff
@@ -334,8 +335,24 @@ pub async fn verify_stack_permissions(
...
+                error!("Tx digest header not found");
+                StatusCode::BAD_REQUEST
+            })?
+            .to_str()
+            .map_err(|_| {
+                error!("Tx digest cannot be converted to a string");
+                StatusCode::BAD_REQUEST
+            })?;
+        let tx_digest = TransactionDigest::from_str(tx_digest_str).unwrap();
+        let compute_units = utils::request_blockchain_for_stack(&state, tx_digest).await?;
```
Cifko, line 351: You should create the stack and deduct the total_num_compute_units from it right away (as it done in the `GetAvailableStackWithComputeUnits`)

Cifko (changes requested):

atoma-service/src/middleware.rs
jorgeantonio21: This makes a lot of sense, thank you !

atoma-service/src/middleware.rs
jorgeantonio21: Great catch !

atoma-service/src/middleware.rs
jorgeantonio21: Addressed

atoma-service/src/middleware.rs
```diff
@@ -334,8 +335,31 @@ pub async fn verify_stack_permissions(
...
+            })?;
+        let tx_digest = TransactionDigest::from_str(tx_digest_str).unwrap();
+        let (tx_stack_small_id, compute_units) =
+            utils::request_blockchain_for_stack(&state, tx_digest, total_num_compute_units).await?;
+
+        // NOTE: We need to check that the stack small id matches the one in the request
+        // otherwise, the user is requesting for a different stack, which is invalid. We
+        // must also check that the compute units are enough for processing the request.
+        if stack_small_id != tx_stack_small_id as i64
+            || compute_units > total_num_compute_units as u64
```
Cifko, line 358: at this stage, the compute units check should already be done.
it should be done at the `insert_new_stack`, and there should be some logic on conflict.

Cifko (changes requested):

atoma-state/src/state_manager.rs
```diff
@@ -1110,7 +1110,14 @@ impl AtomaState {
...
                 VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9, $10, $11)
-                ON CONFLICT (stack_small_id) DO NOTHING",
+                ON CONFLICT (stack_small_id) DO UPDATE
+                SET already_compute_units = 
+                    CASE 
+                        WHEN stacks.already_computed_units + $8 <= stacks.num_compute_units
+                        THEN stacks.already_computed_units + $8
+                        ELSE RAISE EXCEPTION 'Computed units would exceed total compute units'
+                    END
+                WHERE stacks.already_computed_units + $8 <= stacks.num_compute_units",
```
Cifko, line 1120: this is good, one more thing and then we are good
during the first insert you can request have `already_computed_units` greater than `num_compute_units` which is bad
we can solve this by adding `constraint` to the table, and then we could also simplify the update query, because the constraint should take care of that 
the constraint should be `CONSTRAINT check_compute_units CHECK (already_computed_units <= num_compute_units)`

Cifko (changes requested):

atoma-state/src/state_manager.rs
jorgeantonio21: Thanks for the suggestion, I addressed it the way you propose

Cifko (approved):

PR #251, opened 2024-11-24
feat: add atoma node key rotation with intel tdx remote attestation and a tdx service

jorgeantonio21:


atoma-sui/src/client.rs
```diff
@@ -1005,6 +1017,55 @@ impl AtomaSuiClient {
         Ok(response.digest.to_string())
     }
 
+    #[instrument(level = "info", skip_all, fields(
```
Cifko, line 1020: add rustdoc

Cifko (approved):
Can you please add overall description on how the rotation works. What keys are being rotate. What's the shared secret for, etc.

atoma-sui/src/client.rs
jorgeantonio21: Addressed

dayeol (approved):

atoma-sui/src/subscriber.rs
```diff
@@ -446,6 +446,8 @@ fn write_cursor_to_toml_file(cursor: Option<EventID>, path: &str) -> Result<()>
 /// * `StackSettlementTicketEvent` - Handles stack settlement tickets
 /// * `StackSettlementTicketClaimedEvent` - Handles claimed stack settlement tickets
 /// * `StackAttestationDisputeEvent` - Handles stack attestation disputes
+/// * `NewKeyRotationEvent` - Handles new key rotation requests
+/// * `NodeKeyRotationEvent` - Handles node key rotation events
```
dayeol, line 450: question - what is difference between these two?

jorgeantonio21: The event `NewKeyRotationEvent` relates to a new epoch key rotation event, applied to every node.

The event `NodeKeyRotationEvent` is emitted whenever a node submits a transaction with a new generated public key. So that the entire network is aware that the given node has committed to a new public key.

@dayeol does this make sense to you ?

PR #253, opened 2024-11-25
feat: add sender to node subscription event

Cifko:
Add sender to node subscription event. This way we can check signatures on other messages that claims to be from this node (e.g. update node public address )


jorgeantonio21 (approved):
LGTM

PR #254, opened 2024-11-26
chore: move verify signature to atoma-utils

Cifko:
Move the signature verification to atoma-utils, so it can be reused in atoma-proxy


jorgeantonio21 (approved):
LGTM

PR #268, opened 2024-12-05
fix: resolve bugs with confidential compute

jorgeantonio21:


atoma-service/src/middleware.rs
```diff
@@ -488,37 +514,80 @@ pub async fn confidential_compute_middleware(
...
+    })?;
+    let cyphertext = body_json
+        .get(atoma_utils::constants::CYPHERTEXT)
+        .ok_or_else(|| {
+            error!("Cyphertext not found in body");
+            StatusCode::BAD_REQUEST
+        })?
+        .as_array()
+        .ok_or_else(|| {
+            error!("Cyphertext is not a string");
```
Cifko, line 573: is not an array?

atoma-confidential/src/key_management.rs
```diff
@@ -154,10 +178,65 @@ impl X25519KeyPairManager {
...
+    /// Writes the current private key to a file at the root directory.
+    ///
+    /// # Warning
+    /// This function is intended for development/testing purposes only.
+    /// Writing private keys to disk in production is a security risk.
+    ///
+    /// # Returns
+    /// * `Ok(())` if the write was successful
+    /// * `Err(KeyManagementError)` if the write failed
+    pub fn write_private_key_to_file(&self) -> Result<()> {
```
Cifko, line 213: hide behind a feature flag, so when we are going to deploy real code we will not use it by accident
maybe a feature flag like "dont_use_in_production"

Cifko (approved):

atoma-service/src/middleware.rs
jorgeantonio21: Yes ! ahah

atoma-confidential/src/key_management.rs
jorgeantonio21: I think we need to always store the private key

PR #283, opened 2024-12-13
feat: add sui timestamp, to stack create/settle

Cifko:
Add timestamp to stack create/settler, it's for the dashboard.

### DIFF
--- a/atoma-confidential/src/key_management.rs
+++ b/atoma-confidential/src/key_management.rs
@@ -212,6 +212,7 @@ impl X25519KeyPairManager {
     /// * `Err(KeyManagementError)` if the write failed
     pub fn write_private_key_to_file(&self) -> Result<()> {
         use std::fs::{self, create_dir_all};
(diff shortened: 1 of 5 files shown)


atoma-confidential/src/key_management.rs
```diff
@@ -212,6 +212,7 @@ impl X25519KeyPairManager {
     /// * `Err(KeyManagementError)` if the write failed
     pub fn write_private_key_to_file(&self) -> Result<()> {
         use std::fs::{self, create_dir_all};
+        #[cfg(unix)]
```
jorgeantonio21, line 215: why this ?

jorgeantonio21 (approved):
I see now

LGTM

atoma-confidential/src/key_management.rs
Cifko: because I want be able to compile the atoma-node on windows ;-)

PR #285, opened 2024-12-13
feat: register on proxy

Cifko:
Add config settings for the proxy where we want to register. Added the `proxy_server` section
- `proxy_address` address of the proxy where to send the registration
- `node_public_address` that's our address where the proxy can send http requests
- `country` the country of this node (this will be removed if we can get this information from the `node_public_address`

The registration is signed with the sui keystore (don't forget to set the `-a` index when testing locally with more than one sui address)


jorgeantonio21 (approved):
LGTM

PR #286, opened 2024-12-15
feat: logic for atoma p2p for gossip nodes public urls

jorgeantonio21:


atoma-p2p/src/service.rs
```diff
@@ -0,0 +1,1066 @@
...
+    /// * Future timestamp manipulation attempts
+    #[instrument(level = "debug", skip_all)]
+    pub(crate) fn validate_public_address_message(
+        response: &AddressResponse,
+    ) -> Result<(), AtomaP2pNodeError> {
+        let now = std::time::Instant::now().elapsed().as_secs();
+
+        // Check if the URL is valid
+        // TODO: Better validation using URL crate with `parse` method ?
+        if !response.address.starts_with("http://") && !response.address.starts_with("https://") {
```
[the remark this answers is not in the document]
jorgeantonio21: Good observation !

atoma-p2p/src/service.rs
```diff
@@ -0,0 +1,1066 @@
...
+                // Verify the node small ID ownership
+                utils::verify_node_small_id_ownership(
+                    node_small_id,
+                    signature.as_slice(),
+                    self.state_manager_sender.clone(),
+                )
+                .await?;
+                // Rebroadcast the message to the network peers
+                let topic = gossipsub::IdentTopic::new(GOSPUBSUB_TOPIC);
+                if let Err(e) = self
```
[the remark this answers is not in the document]
jorgeantonio21: I removed this logic and added:

```rs
self.swarm
                    .behaviour_mut()
                    .gossipsub
                    .report_message_validation_result(message_id, propagation_source, gossipsub::MessageAcceptance::Accept)?;
```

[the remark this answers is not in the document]
jorgeantonio21: I believe in case the previous validation steps error, the gossipsub will automatically not propagate the message. Or should I enforce it through the `report_message_validation_result` to be `MessageAcceptance::Reject`?

atoma-p2p/src/service.rs
```diff
@@ -0,0 +1,1066 @@
...
+/// Network behavior configuration for the P2P Atoma node, combining multiple libp2p protocols.
+///
+/// This struct implements the `NetworkBehaviour` trait and coordinates three main networking components
+/// for peer discovery, message broadcasting, and distributed routing.
+#[derive(NetworkBehaviour)]
+struct MyBehaviour {
+    /// Handles publish-subscribe messaging across the P2P network.
+    /// Used for broadcasting node addresses and other network messages using a gossip protocol
+    /// that ensures efficient message propagation.
+    gossipsub: gossipsub::Behaviour,
```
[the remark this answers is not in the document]
jorgeantonio21: Thank you, this is very helpful. I added the kademlia and mdns behaviors back to the atoma-p2p-service. I believe `libp2p` internally handles all discovery/updates on the DHT and the subsequent interactions with the `gossipsub` layer. For this reason, I didn't add any further logic for event handling. Let me know if my assumptions are correct, or there needs to be more work done here.

atoma-p2p/src/service.rs
```diff
@@ -0,0 +1,1078 @@
...
+                // To content-address message, we can take the hash of message and use it as an ID.
+                let message_id_fn = |message: &gossipsub::Message| {
+                    let mut s = DefaultHasher::new();
+                    message.data.hash(&mut s);
+                    gossipsub::MessageId::from(s.finish().to_string())
+                };
+
+                let gossipsub_config = gossipsub::ConfigBuilder::default()
+                    .heartbeat_interval(config.heartbeat_interval)
+                    .validation_mode(gossipsub::ValidationMode::Strict)
```
sdbondi, line 190: You need to add validate_message here
https://docs.rs/gossipsub/latest/gossipsub/struct.GossipsubConfig.html#method.validate_messages

atoma-p2p/src/service.rs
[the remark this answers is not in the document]
sdbondi: You always need to provide a resolution to the message - so should explicitly reject in the error case.

atoma-p2p/src/service.rs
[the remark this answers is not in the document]
sdbondi: That's correct, essentially kad takes care of connecting to neighbours and discovering public addresses. These are then available to gossipsub.

atoma-p2p/src/service.rs
jorgeantonio21: Oh nice one, there it goes

atoma-p2p/src/service.rs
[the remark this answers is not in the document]
jorgeantonio21: I added some more logic for this

atoma-p2p/src/service.rs
```diff
@@ -0,0 +1,1125 @@
...
+        }
+        let gossip_message: GossipMessage = serde_cbor::from_slice(message_data)?;
+        match gossip_message {
+            GossipMessage::AddressResponse(response) => {
+                debug!(
+                    target = "atoma-p2p",
+                    event = "gossipsub_message_data",
+                    "Received gossipsub message data"
+                );
+                let message_acceptance =
```
jorgeantonio21, line 547: @sdbondi I added here error handling for message validation, and providing the acceptance criteria resulting from it, which is then passed below in the `report_message_validation_result`. How does this look like ?

atoma-p2p/src/service.rs
```diff
@@ -0,0 +1,1125 @@
...
+                }
+            }
+            GossipMessage::AddressRequest => {
+                debug!(
+                    target = "atoma-p2p",
+                    event = "gossipsub_message_data",
+                    "Received gossipsub message data"
+                );
+                // Address request messages are always accepted and propagated on the p2p network
+                self.swarm
```
jorgeantonio21, line 598: @sdbondi also added a `report_message_validation_result` to always accept requests for node addresses, as if I understand correctly, otherwise this message would not be propagated to other nodes.

sdbondi: That is correct - however since gossipsub gossips the message to all subscribers, you probably don't want to use it for this part of the protocol. i.e. a single request will arrive at all subscribers (i.e all nodes) and will result in all nodes sending a response to all nodes. I assume that all nodes also send a request. $O(n^2)$ message complexity.

https://docs.rs/libp2p-request-response/latest/libp2p_request_response/ might be what you're looking for

atoma-p2p/src/timer.rs
```diff
@@ -0,0 +1,125 @@
...
+    node_small_id: u64,
+    country: String,
+    usage_metrics_tx: UnboundedSender<NodeUsageMetrics>,
+) -> JoinHandle<Result<(), AtomaP2pNodeError>> {
+    tokio::spawn(async move {
+        // NOTE: We only publish usage metrics for nodes, clients do not need to publish any usage metrics
+        if !is_client {
+            loop {
+                let next_heartbeat =
+                    duration_until_next_heartbeat(USAGE_METRICS_HEARTBEAT_INTERVAL);
```
maschad, line 55: This `duration_until_next_heartbeat` seems a bit excessive since this is a heartbeat system, we can just use a fixed interval sleep

```rust
        tokio::time::sleep(USAGE_METRICS_HEARTBEAT_INTERVAL).await;
```

maschad (changes requested):
I'm still not convinced having p2p nodes register their public url is an effective mechanism for [peer routing](https://docs.libp2p.io/concepts/discovery-routing/overview/#peer-routing) 

If we are going to be using Kademlia DHTs for discovery this seems counter intuitive, nodes should be located by their [PeerID](https://docs.libp2p.io/concepts/fundamentals/peers/#peer-id)

It may be useful to examine how [Kademlia](https://pl-launchpad.io/curriculum/libp2p/dht/) works so that this is clearer

atoma-p2p/src/timer.rs
jorgeantonio21: I agree, I will refactor this logic to be simpler and send messages every fixed interval (without any additional node synchronization).

atoma-p2p/src/config.rs
```diff
@@ -0,0 +1,94 @@
...
+    pub public_url: String,
+
+    /// The node's small id (assigned by the Atoma smart contract)
+    pub node_small_id: u64,
+
+    /// The list of seed nodes to connect to.
+    ///
+    /// Seed nodes are nodes that the node will use to bootstrap its network connection.
+    /// They are a list of strings in the format of "/ip4/x.x.x.x/tcp/x".
+    pub seed_nodes: Vec<String>,
```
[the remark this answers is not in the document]
jorgeantonio21: This makes a lot of sense. Are these valid bootstrap that we can leverage ?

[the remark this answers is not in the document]
maschad: Yes they are, they are well connected nodes so it gives us access to an already rich DHT

atoma-p2p/src/service.rs
```diff
@@ -0,0 +1,1138 @@
...
+                target = "atoma-p2p",
+                event = "gossipsub_message_from_self",
+                "Gossipsub message from self"
+            );
+            // Do not re-publish the node's own message, just return `Ok(())
+            return Ok(());
+        }
+        let gossip_message: GossipMessage = ciborium::from_reader(message_data)?;
+        match gossip_message {
+            GossipMessage::UsageMetrics(usage_metrics) => {
```
maschad, line 571: Do we intend to gossip other types of messages for the time being? Otherwise I don't think `UsageMetrics` needs to be an enum here, we can just use the type as done [here](https://github.com/atoma-network/atoma-node/pull/286/files#diff-f558264be9dc74d18752e874e9306be2942281ddfc84762ae59c17ae15b74d64R730)

atoma-p2p/src/service.rs
```diff
@@ -0,0 +1,1138 @@
...
+    #[error("Failed to subscribe to topic: {0}")]
+    GossipsubSubscriptionError(#[from] gossipsub::SubscriptionError),
+    #[error("Failed to listen on address: {0}")]
+    SwarmListenOnError(#[from] TransportError<std::io::Error>),
+    #[error("Failed to parse listen address: {0}")]
+    ListenAddressParseError(#[from] libp2p::multiaddr::Error),
+    #[error("Failed to dial bootstrap node: {0}")]
+    BootstrapNodeDialError(#[from] DialError),
+    #[error("Failed to sign hashed message: {0}")]
+    SignHashedMessageError(String),
```
maschad, line 775: How's this different than the [`SignatureError`](https://github.com/atoma-network/atoma-node/pull/286/files#diff-f558264be9dc74d18752e874e9306be2942281ddfc84762ae59c17ae15b74d64R787) ?

atoma-p2p/src/service.rs
```diff
@@ -0,0 +1,1138 @@
...
+        Ok(())
+    }
+}
+
+#[derive(Debug, Error)]
+pub enum AtomaP2pNodeError {
+    #[error("Failed to build gossipsub config: {0}")]
+    GossipsubConfigError(#[from] ConfigBuilderError),
+    #[error("Failed to build gossipsub: {0}")]
+    GossipsubBuildError(String),
```
maschad, line 761: These two errors seems to address the same thing

atoma-p2p/src/service.rs
```diff
@@ -0,0 +1,1138 @@
...
+    /// # Arguments
+    ///
+    /// * `response` - The message received from the P2P network
+    /// * `state_manager_sender` - The sender of the state manager
+    ///
+    /// # Returns
+    ///
+    /// Returns `Ok(())` if the message is valid, or a `AtomaP2pNodeError` if any step fails.
+    #[instrument(level = "debug", skip_all)]
+    pub async fn validate_usage_metrics_message(
```
maschad, line 1108: Let's an integration test for this given it's a combination of all of the functions

maschad (changes requested):
Awesome work on this @jorgeantonio21 🚀 

Looking good but left some comments

atoma-p2p/src/config.rs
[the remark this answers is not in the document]
jorgeantonio21: If other networks are using them for bootstrapping, how can they only bootstrap for our nodes/clients ? Through gossip sub topic (I imagine there is no uniqueness requirements for it).

atoma-p2p/src/service.rs
```diff
@@ -0,0 +1,1138 @@
...
+    #[instrument(level = "info", skip_all)]
+    pub fn start(
+        config: AtomaP2pNodeConfig,
+        keystore: Arc<FileBasedKeystore>,
+        state_manager_sender: Sender<StateManagerEvent>,
+        is_client: bool,
+    ) -> Result<Self, AtomaP2pNodeError> {
+        let mut swarm = SwarmBuilder::with_new_identity()
+            .with_tokio()
+            .with_tcp(
```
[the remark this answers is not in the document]
jorgeantonio21: Yeah, I believe so too. I was wondering if we would need in extreme scenarios (where nodes don't support `QUIC` for some reason) and could fallback to `TCP`. But I agree, let's move with only `QUIC`

atoma-p2p/src/service.rs
```diff
@@ -0,0 +1,1138 @@
...
+                            debug!(
+                                target = "atoma-p2p",
+                                event = "gossipsub_unsubscribed",
+                                peer_id = %peer_id,
+                                topic = %topic,
+                                "Peer unsubscribed from topic"
+                            );
+                            // TODO: should we remove the locally stored public url from an unsubscribed peer?
+                        }
+                        _ => {}
```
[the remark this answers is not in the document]
jorgeantonio21: Are all these the events to handle ? If I understand correctly gossipsub already integrates kademlia for peer discovery, so there seems to not be much to do on that front..

atoma-p2p/src/config.rs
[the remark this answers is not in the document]
jorgeantonio21: Also, I doubt these nodes will subscribe to our gossipsub topics, so I am unsure this will be of any help for now.

atoma-p2p/src/service.rs
jorgeantonio21: Yes, I do believe we want to add other gossip message types

atoma-p2p/src/service.rs
jorgeantonio21: Addressed

atoma-p2p/src/config.rs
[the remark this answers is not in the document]
maschad: Before a peer can subscribe to a topic it must find other peers and establish network connections with them. The pub/sub system doesn’t have any way to discover peers by itself.  The main purpose of boostrapping is to gain an overview of the network to accelerate peer discovery, gossipsub is not a peer discovery mechanism.

atoma-p2p/src/service.rs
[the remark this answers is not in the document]
maschad: That's inaccurate, one has to explicitly add them when the `Discovered` event is emitted as you can see the [discovered_nodes](https://github.com/libp2p/rust-libp2p/blob/master/protocols/mdns/src/behaviour.rs#L180) is not merged into the gossipsub nor Kademlia properties, which is why in this [example](https://github.com/libp2p/rust-libp2p/blob/master/examples/distributed-key-value-store/src/main.rs#L93) it's done explicitly

atoma-p2p/src/service.rs
```diff
@@ -0,0 +1,1404 @@
...
+    /// * Replay attacks by enforcing timestamp freshness
+    /// * Future timestamp manipulation attempts
+    #[instrument(level = "debug", skip_all)]
+    pub fn validate_usage_metrics_url_and_timestamp(
+        usage_metrics: &UsageMetrics,
+    ) -> Result<(), AtomaP2pNodeError> {
+        let now = std::time::Instant::now().elapsed().as_secs();
+
+        // Check if the URL is valid
+        let _usage_metrics_url = Url::parse(&usage_metrics.node_public_url).map_err(|e| {
```
maschad, line 925: I don't think it's sufficient to check if the url has a valid format, I think for now we should have an allow list of set urls providers who can join the mesh and report and remove any peers who have URLs outside that which try to subscribe. This can later be revisited as we define more rigid [peer scoring](https://github.com/libp2p/specs/blob/master/pubsub/gossipsub/gossipsub-v1.1.md#peer-scoring)

atoma-p2p/src/service.rs
```diff
@@ -0,0 +1,1404 @@
...
+                )?;
+
+                let store = kad::store::MemoryStore::new(key.public().to_peer_id());
+                let kademlia = kad::Behaviour::new(key.public().to_peer_id(), store);
+
+                let mdns = mdns::tokio::Behaviour::new(
+                    mdns::Config::default(),
+                    key.public().to_peer_id(),
+                )?;
+
```
maschad, line 203: We should also include the [identify behaviour](https://github.com/libp2p/rust-libp2p/blob/56691dbc33de82e221da14ccea5edc04074e94c4/examples/identify/README.md#L4) as a part of the config

atoma-p2p/src/service.rs
jorgeantonio21: That's a great idea, I have created an issue for this. So that we can address it later

jorgeantonio21: See #390

maschad (commented):
LGTM minus this one comment https://github.com/atoma-network/atoma-node/pull/286#discussion_r1945153329 Great job @jorgeantonio21  🚀

maschad (approved):
Excellent work!

PR #287, opened 2024-12-16
feat: add prometheus docker service for node performance metrics analytics

jorgeantonio21:


docker-compose.yaml
francis2tm, line 1: Does `atoma-node` need to wait for prometheus or grafana to start? Like it does for the inference services

francis2tm (commented):
LGTM, just one comment

docker-compose.yaml
jorgeantonio21: I think the right flow, would be for prometheus and grafana to wait for the node to start, no ? However, for now maybe we want to leave it without dependencies, so that anyone can deploy the containers and run the node manually..

francis2tm (approved):

Issue #274, opened 2024-12-06
We redo the shared secret computation, might be best to stored it in memory for later use


jorgeantonio21:
Ideally, we would benchmark both approaches.

PR #289, opened 2024-12-17
feat: remove file storing for symmetric key and submit rotate key on node start

jorgeantonio21:
Addresses #274

### DIFF
--- a/atoma-bin/atoma_node.rs
+++ b/atoma-bin/atoma_node.rs
@@ -227,17 +227,21 @@ async fn main() -> Result<()> {
 
     let (compute_shared_secret_sender, compute_shared_secret_receiver) =
         tokio::sync::mpsc::unbounded_channel();
-    let confidential_compute_service = AtomaConfidentialComputeService::new(
-        client.clone(),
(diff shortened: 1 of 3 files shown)


Cifko (approved):

atoma-bin/atoma_node.rs
```diff
@@ -227,17 +227,21 @@ async fn main() -> Result<()> {
...
-        app_state_encryption_receiver,
-        compute_shared_secret_receiver,
-        shutdown_receiver.clone(),
-    )?;
 
+    let client_clone = client.clone();
+    let shutdown_receiver_clone = shutdown_receiver.clone();
     spawn_with_shutdown(
-        async move { confidential_compute_service.run().await },
+        async move {
```
Cifko, line 234: we don't need the async move, remove that and then you can get rid of the two clones and do them inline

Cifko (approved):

atoma-bin/atoma_node.rs
jorgeantonio21: Oh nice ! Thanks for the insight

Issue #291, opened 2024-12-19
CC service spawns and a key rotation is triggered, we need to store the key rotation counter, and only update the keys for larger counters


jorgeantonio21:

jorgeantonio21:
Addressed in #292

PR #292, opened 2024-12-19
feat: add logic to handle node public key submissions only starting from the current key rotation counter

jorgeantonio21:


atoma-bin/atoma_node.rs
```diff
@@ -225,16 +226,17 @@ async fn main() -> Result<()> {
         register_on_proxy(&config.proxy, *node_small_id, &keystore, args.address_index).await?;
     }
 
-    let (compute_shared_secret_sender, compute_shared_secret_receiver) =
+    let (compute_shared_secret_sender, _compute_shared_secret_receiver) =
         tokio::sync::mpsc::unbounded_channel();
 
+    #[cfg(feature = "tdx")]
     spawn_with_shutdown(
```
Cifko, line 233: you should keep the `handle` that's returned from the spawn_with_shutdown and put that into the `try_join` at L380.

atoma-bin/atoma_node.rs
```diff
@@ -225,16 +226,17 @@ async fn main() -> Result<()> {
         register_on_proxy(&config.proxy, *node_small_id, &keystore, args.address_index).await?;
     }
 
-    let (compute_shared_secret_sender, compute_shared_secret_receiver) =
+    let (compute_shared_secret_sender, _compute_shared_secret_receiver) =
         tokio::sync::mpsc::unbounded_channel();
 
+    #[cfg(feature = "tdx")]
```
Cifko, line 232: if we want to hide this behind a feature flag, maybe we can put all TDX stuff behind a feature flag, that will reduce the size/compilation time when building without it. But I'm not sure how the code acts when it doesn't have confidentiality support but it receives confidential compute request anyway, what happens than? Maybe we can put everything behind feature flag and just don't expose the endpoint at all. Any idea what's better?

Cifko (approved):
LGTM just add the handle the other comment doesn't need to resolve into complex more PR ;-)

atoma-bin/atoma_node.rs
jorgeantonio21: I see, applied changes.

Cifko (approved):

jorgeantonio21:
Addresses #291

PR #297, opened 2024-12-20
feat: review atoma daemon api docs and do not panic when connecting to proxy server (at spawn)

jorgeantonio21:


atoma-bin/atoma_node.rs
```diff
@@ -213,6 +213,17 @@ async fn main() -> Result<()> {
...
         tokio::sync::mpsc::unbounded_channel();
 
+    for (_, node_small_id) in config.daemon.node_badges.iter() {
+        if let Err(e) = register_on_proxy(&config.proxy, *node_small_id, &keystore, args.address_index).await {
+            error!(
+                target = "atoma-node-service",
+                event = "register_on_proxy_error",
+                error = ?e,
+                "Failed to register on proxy server"
+            );
```
Cifko, line 223: should we exit the application? this means the proxy doesn't know the address of the node, so it can't call the endpoints

jorgeantonio21: I honestly don't know what's best. This overall approach is not good, we need a p2p network for public url sharing..

Cifko: yeah, agreed

Cifko (approved):

PR #319, opened 2024-12-31
feat: refactor confidential compute endpoints

jorgeantonio21:


atoma-service/src/components/openapi.rs
```diff
@@ -16,13 +24,19 @@ pub fn openapi_routes() -> Router {
...
+            (path = CONFIDENTIAL_EMBEDDINGS_PATH, api = ConfidentialEmbeddingsOpenApi),
+            (path = CONFIDENTIAL_CHAT_COMPLETIONS_PATH, api = ConfidentialChatCompletionsOpenApi),
         ),
         tags(
             (name = "health", description = "Health check"),
             (name = "metrics", description = "Metrics"),
             (name = "chat", description = "Chat completions"),
             (name = "embeddings", description = "Embeddings"),
             (name = "images", description = "Image generations"),
+            (name = "confidential-images", description = "Confidential image generations"),
```
Cifko, line 37: we are not actually using the tags. Also we should use them, but not like this. Doesn't make any sense tag equal function. They should be used properly, grouping similar function. But it doesn't have to be part of this PR.

atoma-service/src/handlers/embeddings.rs
```diff
@@ -116,6 +117,115 @@ pub async fn embeddings_handler(
...
+    fields(path = request_metadata.endpoint_path)
+)]
+pub async fn confidential_embeddings_handler(
+    Extension(request_metadata): Extension<RequestMetadata>,
+    State(state): State<AppState>,
+    Json(payload): Json<Value>,
+) -> Result<Json<Value>, AtomaServiceError> {
+    info!("Received embeddings request, with payload: {payload}");
+    let model = payload
+        .get("model")
```
Cifko, line 184: in the other file you introduced `STREAM_KEY`. Maybe it would be nice to have `MODEL_KEY` as well.

atoma-service/src/handlers/image_generations.rs
```diff
@@ -119,6 +120,115 @@ pub async fn image_generations_handler(
...
+    fields(path = request_metadata.endpoint_path)
+)]
+pub async fn confidential_image_generations_handler(
+    Extension(request_metadata): Extension<RequestMetadata>,
+    State(state): State<AppState>,
+    Json(payload): Json<Value>,
+) -> Result<Json<Value>, AtomaServiceError> {
+    info!("Received image generations request, with payload: {payload}");
+    let model = payload
+        .get("model")
```
Cifko, line 185: same here, `MODEL_KEY`

atoma-service/src/streamer.rs
```diff
@@ -33,15 +34,34 @@ use crate::{
...
+const NONCE_KEY: &str = "nonce";
+
+/// The response hash key
+const RESPONSE_HASH_KEY: &str = "response_hash";
+
+/// The signature key
+const SIGNATURE_KEY: &str = "signature";
+
+/// The usage key
+const USAGE_KEY: &str = "usage";
```
Cifko, line 63: You have the same keys in the atoma-service/src/handlers/mod.rs, maybe move it to a new file and use from there.

atoma-service/src/streamer.rs
```diff
@@ -176,23 +196,12 @@ impl Streamer {
             payload_hash = hex::encode(self.payload_hash)
         )
     )]
-    fn handle_final_chunk(&mut self, usage: &Value) -> Result<String, Error> {
+    fn handle_final_chunk(&mut self, usage: &Value, response_hash: [u8; 32]) -> Result<(), Error> {
```
Cifko, line 199:
```suggestion
    fn handle_final_chunk(&mut self, usage: &Value, response_hash: [u8; PAYLOAD_HASH_SIZE]) -> Result<(), Error> {
```

atoma-service/src/middleware.rs
```diff
@@ -1090,4 +1024,267 @@ pub(crate) mod utils {
...
+            })?;
+        let salt_bytes: [u8; SALT_SIZE] = salt_bytes.try_into().map_err(|e| {
+        AtomaServiceError::InvalidHeader {
+            message: format!(
+                "Failed to convert salt bytes to {SALT_SIZE}-byte array, incorrect length, with error: {:?}",
+                e
+            ),
+            endpoint: endpoint.to_string(),
+        }
+    })?;
```
Cifko, line 1145: formatting?

Cifko (changes requested):

atoma-service/src/handlers/embeddings.rs
jorgeantonio21: Makes sense

atoma-service/src/handlers/image_generations.rs
jorgeantonio21: Resolved

atoma-service/src/middleware.rs
jorgeantonio21: Oh great catch !

Cifko (approved):

PR #320, opened 2025-01-01
feat: resolve some build issues when --features tdx is on

jorgeantonio21:
Also adds an option to build with feature `tdx`, through docker-compose

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -34,12 +34,26 @@ jobs:
           username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
 
-      - name: Build and push Docker image
+      - name: Build and push TDX-enabled Docker image
         uses: docker/build-push-action@v5
         with:
           context: .
           push: true
-          platforms: linux/amd64,linux/arm64
+          platforms: linux/amd64
+          build-args:
+            ENABLE_TDX=true
           tags: |
             ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:latest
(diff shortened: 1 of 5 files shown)


.github/workflows/release.yml
```diff
@@ -34,12 +34,26 @@ jobs:
...
             ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:${{ github.event.release.tag_name }}
+
+    - name: Build and push default Docker image
+        uses: docker/build-push-action@v5
+        with:
+          context: .
+          push: true
+          platforms: linux/amd64, linux/arm64
+          build-args:
+            ENABLE_TDX: "false"
```
Cifko, line 56:
```suggestion
            ENABLE_TDX=false
```

atoma-confidential/src/service.rs
```diff
@@ -255,7 +258,7 @@ impl AtomaConfidentialComputeService {
             let tdx_quote = get_compute_data_attestation(&public_key_bytes)?;
             let tdx_quote_bytes = tdx_quote.to_bytes();
             match self
-                .sui_client
+                ._sui_client
```
Cifko, line 261: why is it `_sui_client` when you are actually using it?

Cifko (approved):

atoma-confidential/src/service.rs
jorgeantonio21: Because it is behind a feature

jorgeantonio21: Disabled dead code

PR #331, opened 2025-01-06
create test coverage script

fishonamos:


.env.example
```diff
@@ -18,6 +18,9 @@ SUI_CONFIG_PATH=~/.sui/sui_config
 # Atoma Node Service Configuration
 ATOMA_SERVICE_PORT=3000
 
+ #Test database
+DATABASE_URL=postgresql://atoma:atoma@localhost/atoma
```
jorgeantonio21, line 22: I would not add these variables here

.env.example
```diff
@@ -18,6 +18,9 @@ SUI_CONFIG_PATH=~/.sui/sui_config
...
+RUST_BACKTRACE=1 # enable backtrace
```
jorgeantonio21, line 23: Same as above

maschad:
Closing this as we [now use codecov](https://github.com/atoma-network/atoma-node/pull/381#issuecomment-2622878625)

PR #346, opened 2025-01-20
feat: add metrics to track successful compute requests as well as int…

maschad:
…ra token times

### DIFF
--- a/atoma-service/src/handlers/chat_completions.rs
+++ b/atoma-service/src/handlers/chat_completions.rs
@@ -144,6 +144,7 @@ pub async fn chat_completions_handler(
         estimated_total_compute_units,
         payload_hash,
         client_encryption_metadata,
+        endpoint_path,
         ..
     } = request_metadata;
     info!(
@@ -153,12 +154,16 @@ pub async fn chat_completions_handler(
         "Received chat completions request, with payload hash: {payload_hash:?}"
     );
 
-    // Check if streaming is requested
     let is_stream = payload
         .get(STREAM_KEY)
         .and_then(|s| s.as_bool())
         .unwrap_or_default();
-    let endpoint = request_metadata.endpoint_path.clone();
+    let endpoint = endpoint_path.clone();
+
+    let client_public_key = client_encryption_metadata
+        .as_ref()
+        .map(|m| hex::encode(m.client_x25519_public_key))
+        .unwrap_or_default();
 
     match handle_response(
         &state,
@@ -172,8 +177,16 @@ pub async fn chat_completions_handler(
     )
     .await
     {
-        Ok(response) => Ok(response),
+        Ok(response) => {
+            TOTAL_COMPLETED_REQUESTS
+                .with_label_values(&[&client_public_key])
+                .inc();
+            Ok(response)
+        }
         Err(e) => {
+            TOTAL_FAILED_REQUESTS
+                .with_label_values(&[&client_public_key])
+                .inc();
             // NOTE: We need to update the stack number of tokens as the service failed to generate
             // a proper response. For this reason, we set the total number of tokens to 0.
             // This will ensure that the stack number of tokens is not updated, and the stack
@@ -187,7 +200,7 @@ pub async fn chat_completions_handler(
             )?;
             return Err(AtomaServiceError::InternalError {
                 message: format!("Error handling chat completions response: {}", e),
-                endpoint: request_metadata.endpoint_path.clone(),
+                endpoint: endpoint.clone(),
             });
         }
     }
@@ -297,6 +310,7 @@ pub async fn confidential_chat_completions_handler(
         estimated_total_compute_units,
         payload_hash,
         client_encryption_metadata,
+        endpoint_path,
         ..
     } = request_metadata;
     info!(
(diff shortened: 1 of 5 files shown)


atoma-service/src/handlers/chat_completions.rs
```diff
@@ -153,12 +154,16 @@ pub async fn chat_completions_handler(
...
 
-    // Check if streaming is requested
     let is_stream = payload
         .get(STREAM_KEY)
         .and_then(|s| s.as_bool())
         .unwrap_or_default();
-    let endpoint = request_metadata.endpoint_path.clone();
+    let endpoint = endpoint_path.clone();
+
+    let client_public_key = client_encryption_metadata
```
maschad, line 163: I opted to use the client's public key here as a unique identifier to track successful / failed requests (as opposed to the model given multiple nodes may run the same model)  but I think we may want to also track this with some kind of public ULID per node as well. Thoughts?

jorgeantonio21: I would say that is best to just use `model`, the reason being these are metrics for each node individually. So we might just decouple this in terms of model alone.

atoma-service/src/handlers/chat_completions.rs
```diff
@@ -621,6 +647,10 @@ async fn handle_streaming_response(
         .with_label_values(&[model])
         .start_timer();
 
+    let intra_token_generation_timer = CHAT_COMPLETIONS_INTRA_TOKEN_GENERATION_TIME
```
jorgeantonio21, line 650: For streaming we are already tracking time to first token (TTFT). We do not want to include that value in this measure, the reason being because the time to first token is actually measuring the time the LLM takes to parse the whole prompt (__prefill__ phase), which might contain 1000s of tokens in a single go (all processed in parallel). 

On the other hand, intra token token should happen in the __decoding__ phase, that is, you process each token individually and sequentially (as opposed to the __prefill__ phase). So I suggest starting this timer only after the time to first token has been finished (in the `streamer`).

atoma-service/src/handlers/embeddings.rs
```diff
@@ -94,6 +97,11 @@ pub async fn embeddings_handler(
         .with_label_values(&[model])
         .start_timer();
 
+    let client_public_key = client_encryption_metadata
```
jorgeantonio21, line 100: Again, I suggest just using the `model` as label

atoma-service/src/handlers/image_generations.rs
```diff
@@ -106,8 +114,16 @@ pub async fn image_generations_handler(
     )
     .await
     {
-        Ok(response) => Ok(response),
+        Ok(response) => {
+            TOTAL_COMPLETED_REQUESTS
```
jorgeantonio21, line 118: I suggest using `model` as the sole label in here

atoma-service/src/streamer.rs
```diff
@@ -563,8 +567,19 @@ impl Streamer {
...
+        let intra_token_generation_timer = CHAT_COMPLETIONS_INTRA_TOKEN_GENERATION_TIME
+            .with_label_values(&[&self.model])
+            .start_timer();
+        self.intra_token_generation_timer = Some(intra_token_generation_timer);
+
         let choices = match chunk.get(CHOICES).and_then(|choices| choices.as_array()) {
-            Some(choices) => choices,
+            Some(choices) => {
+                // Observe the intra-token generation timer
+                if let Some(timer) = self.intra_token_generation_timer.take() {
```
jorgeantonio21, line 578: I would observe the timer at the end of the logical flow for the chunks (slightly down the line)

jorgeantonio21 (commented):
Looking good, left a few minor comments

atoma-service/src/handlers/chat_completions.rs
maschad: Makes sense given a node can run multiple models, thanks for clarifying

updated in a383ece

atoma-service/src/handlers/image_generations.rs
maschad: a383ece

atoma-service/src/streamer.rs
maschad: Updated in 540c294

I think it's reasonable to assume this metric excludes the time taken to handle the encryption request for each chunk

atoma-service/src/handlers/chat_completions.rs
maschad: Good point, I've updated in 540c294

atoma-service/src/streamer.rs
jorgeantonio21: I would add it though, as we want to have detailed metrics on total time to generate next token, and if running with confidential compute mode on, encryption/decryption should be taken in consideration

maschad: That's a valid point but I think we should perhaps track that separately to see if encryption requests are a potential bottleneck, token generation from from my perspective should be viewed as a separate process.

30fcf2f

atoma-service/src/streamer.rs
```diff
@@ -563,6 +572,10 @@ impl Streamer {
 
         let (signature, response_hash) = self.sign_chunk(&chunk)?;
 
+        let intra_token_generation_timer = CHAT_COMPLETIONS_INTRA_TOKEN_GENERATION_TIME
```
jorgeantonio21, line 575: I realized the `intra_token_generation_timer` should not be instantiated in here. Instead, it should be at the end of the current logical flow (current stream poll), when the stream polls next then we should observe time and instantiate a new intra token timer.

jorgeantonio21 (commented):
Check the comment I left above

atoma-service/src/streamer.rs
maschad: That makes sense, I've updated it in 7b55718

atoma-service/src/streamer.rs
maschad: Sure, I can agree to that. I've removed the encryption related metrics and updated the PR 7b55718

jorgeantonio21 (approved):
LGTM

PR #348, opened 2025-01-21
feat: country code is required

Cifko:
Add country requirement as country code ISO 3166-1 alpha-2.


config.example.toml
```diff
@@ -41,5 +41,5 @@ node_badges = [
 proxy_address = ""
 # replace this with the public url address of this node
 node_public_address = ""
-# replace this with the country of the node
-country = ""
\ No newline at end of file
+# required, replace this with the country (ISO 3166-1 alpha-2) of the node
```
jorgeantonio21, line 44: Please add a link that one can use to get the right code for the country

jorgeantonio21 (approved):
LGTM

Please add the request link

config.example.toml
Cifko: added

PR #352, opened 2025-01-24
feat: amd-sev-snp draft and feature flags for confidential computing

bredamatt:
TODO:
- []: Add AMD SEV-SNP module
- []: Test AMD SEV-SNP module


atoma-confidential/src/sev_snp.rs
```diff
@@ -0,0 +1,91 @@
...
+pub fn get_compute_data_attestation(attested_data: &[u8]) -> Result<AttestationReport> {
+    if attested_data.len() != SEV_SNP_REPORT_DATA_SIZE {
+        return Err(SevError::InvalidAttestedDataSize(attested_data.len()));
+    }
+
+    let mut firmware = Firmware::open().map_err(SevError::FailedToOpenFirmware)?;
+
+    // Ask the PSP to generate an attestation report for the given data
+    let (attestation_report, cert_chain): (AttestationReport, Vec<CertTableEntry>) = firmware
+        .get_ext_report(Some(1), Some(attested_data), Some(1))
```
jorgeantonio21, line 55: Please add documentation on why the current choice for the `message_version` and the `vmpl`.

atoma-confidential/src/sev_snp.rs
```diff
@@ -0,0 +1,91 @@
...
+        .map_err(SevError::FailedToGetAttestationReport)?;
+
+    // Verify the attestation report signature using the VCEK.
+    // This uses the crypto_nossl feature of the sev crate for detailed error handling and a pure Rust implementation.
+    (attestation_report, cert_chain)
+        .verify()
```
jorgeantonio21, line 61: I don't think we need to verify the attestation report signature at this point. This is better done at the client level

jorgeantonio21 (commented):
Looking good.

Left a few comments

atoma-confidential/src/sev_snp.rs
bredamatt: When you say client, do you mean the atom-node on the same guest, or an atoma-node on a separate machine?
The reason why I included the verification is because the VCEK is only available on the same machine, and used for self-verfiying purposes (i.e. prove that the guest the AttestationReport was created from is a valid SEV-SNP guest since its signed with the VCEK, derived from the AMD Root key). I think the only way to verify that on a separate machine would be to also include the certificate chain, but I am not sure this is desireable.

https://docs.rs/sev/5.0.0/src/sev/certs/snp/mod.rs.html#40-49

atoma-confidential/src/sev_snp.rs
bredamatt: These are just defaults used by the instruction. Will add some more docs.

atoma-confidential/src/sev_snp.rs
jorgeantonio21: Yeah, I mean the atoma client (which can be running on my local machine) should verify attestations that are submitted on-chain, so that it knows if it can trust a given node or not.

atoma-sui/src/events.rs
```diff
@@ -839,6 +839,11 @@ pub struct NodePublicKeyCommittmentEvent {
     /// The TEE remote attestation report attesting for
     /// the public key's generation integrity, in byte format.
     pub tee_remote_attestation_bytes: Vec<u8>,
+
+    /// The TEE Provider used to generate the public key.
+    /// This type is used to identify which attestation
+    /// verification method should be used.
+    pub tee_provider: Vec<u8>
```
jorgeantonio21, line 846: can it be a `u16` or `u8`?

atoma-state/src/migrations/20241202121152_node_public_key_rotation.sql
```diff
@@ -2,6 +2,7 @@
 CREATE TABLE IF NOT EXISTS node_public_key_rotations (
     node_small_id            BIGINT  PRIMARY KEY,
     public_key_bytes         BYTEA    UNIQUE NOT NULL,
-    tdx_quote_bytes          BYTEA    NOT NULL,
-    epoch          BIGINT  NOT NULL
+    tee_quote_bytes          BYTEA    NOT NULL,
+    tee_provider             BYTEA    NOT NULL,
```
jorgeantonio21, line 6: Usually it is a bad practice to change migrations files in place, instead I suggest adding a new migration with a `ALTER TABLE node_public_key_rotations` command

atoma-confidential/src/sev_snp.rs
```diff
@@ -0,0 +1,138 @@
...
+/// # Example
+///
+/// ```
+/// let data_to_attest = [1u8; 64];
+/// match get_compute_data_attestation(&data_to_attest) {
+///     Ok(report) => println!("Attestation successful"),
+///     Err(e) => eprintln!("Attestation failed: {}", e),
+/// }
+/// ```
+pub fn get_compute_data_attestation(attested_data: &[u8]) -> Result<SNPAttestationReport> {
```
jorgeantonio21, line 45: This looks great to me, we do need to look further how to incorporate NVIDIA cc into these remote attestation generation.

jorgeantonio21 (commented):
Looking great,

Left a few more comments in the current logic

atoma-sui/src/events.rs
bredamatt: Should be doable, yep

atoma-confidential/src/sev_snp.rs
bredamatt: Yes, nvtrust port in progress.

jorgeantonio21 (commented):
Looks great overall. Let's wait until we have it tested and then we can merge :)

atoma-state/src/migrations/20250206184237_node_public_key_rotation.sql
```diff
@@ -0,0 +1,8 @@
+-- Create tasks table
```
jorgeantonio21, line 1: This comment is not within context.

atoma-state/src/migrations/20250206184237_node_public_key_rotation.sql
```diff
@@ -0,0 +1,8 @@
...
+CREATE TABLE IF NOT EXISTS node_public_key_rotations (
```
jorgeantonio21, line 2: I suggest instead of re-creating the `node_public_key_rotations` table, that was already created in a previous migrations, to instead alter it (in place). 

Notice that you are creating the table `if not exists`, which already does, so your changes are not applied and you get a bunch of missing column related errors in the tests. 

Instead, I would change the code to something along the lines:

```rs
ALTER TABLE node_public_key_rotations
    DROP COLUMN tdx_quote_bytes,
    ADD COLUMN tee_quote_bytes BYTEA NOT NULL,
    ADD COLUMN tee_provider BYTEA NOT NULL;
```

jorgeantonio21 (commented):
Left some comments, which hopefully will help with the tests.

PR #353, opened 2025-01-24
docs: update `README` with further instructions for setup

maschad:
These changes are minor markdown edits + the results of my experiencing setting up a running docker instance of an Atoma node

Depends on https://github.com/atoma-network/atoma-contracts/pull/132

### DIFF
--- a/README.md
+++ b/README.md
@@ -140,7 +140,7 @@ country = ""
 mkdir -p data logs
 ```
 
-1. Start the containers with the desired inference services, please note if you don't have a GPU, you'll need to use the you will need to use a `mistral` backend as currently the other vllms are not supported on CPU.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -130,15 +134,15 @@ node_public_address = ""
...
 
-4. Create required directories
+1. Create required directories
 
 ```bash
 mkdir -p data logs
 ```
 
-5. Start the containers with the desired inference services
+1. Start the containers with the desired inference services, please note if you don't have a GPU, you'll need to use the you will need to use a `mistral` backend as currently the other vllms are not supported on CPU.
```
jorgeantonio21, line 143: We do have a `vllm-cpu` available container, but it only works with x86_64 archs, so for arm based machines we recommend using `mistralrs-cpu`.

jorgeantonio21 (commented):
LGTM

Left a minor comment to be addressed

README.md
maschad: Thanks for the clarification. I've updated  in 5f595ea

jorgeantonio21 (approved):
LGTM

PR #355, opened 2025-01-26
feat: add support for multiple chat completions services

jorgeantonio21:


atoma-service/src/handlers/chat_completions.rs
```diff
@@ -1206,10 +1213,21 @@ pub(crate) mod utils {
         endpoint: &str,
     ) -> Result<Value, AtomaServiceError> {
         let client = Client::new();
+        let model = payload.get("model").unwrap().as_str().unwrap();
```
Cifko, line 1216:
```suggestion
        let model = payload.get(MODEL_KEY).unwrap().as_str()?;
```

Cifko (approved):

atoma-service/src/handlers/chat_completions.rs
jorgeantonio21: Makes sense, added an `unknown` key

francis2tm (approved):
LGTM

PR #356, opened 2025-01-27
ci: add dependabot

maschad:
We can use [dependabot](https://github.com/dependabot) to easily manage deps


jorgeantonio21 (approved):
Love this !

LGTM

PR #368, opened 2025-01-27
ci: add some dependencies to be ignored

maschad:
We currently want to update these deps manually if we choose to do so until we have more solid end-to-end tests


jorgeantonio21 (approved):
Thanks

LGTM

PR #369, opened 2025-01-27
build: fix issue related to rustls audit

maschad:
Our security audit detected [a vulnerability](https://github.com/atoma-network/atoma-node/actions/runs/12998894498/job/36253012065) in `rustls` which is a transitive dependency. 

This applies a patch for it.

The other current [RSA timing vulnerability](https://rustsec.org/advisories/RUSTSEC-2023-0071.html)  is currently unpatched but doesn't apply to us.


jorgeantonio21 (approved):
LGTM

PR #378, opened 2025-01-29
feat: deprecate the use of max_tokens to max_completion_tokens, following openai api spec

jorgeantonio21:
Deprecate the use of `max_tokens` to `max_completion_tokens`, following OpenAI's API spec


maschad (approved):
LGTM

PR #381, opened 2025-01-29
ci: add workflow dispatch for e2e tests

maschad:
In order to run these tests manually I've configured it to run on the `workflow_dispatch` event.

For more info see https://docs.github.com/en/actions/managing-workflow-runs-and-deployments/managing-workflow-runs/manually-running-a-workflow


jorgeantonio21 (approved):
LGTM

PR #392, opened 2025-02-06
chore: add more metrics for alerts on failed requests

maschad:


atoma-service/src/handlers/chat_completions.rs
```diff
@@ -354,6 +365,9 @@ pub async fn confidential_chat_completions_handler(
             Ok(response)
         }
         Err(e) => {
+            TOTAL_FAILED_CHAT_CONFIDENTIAL_REQUESTS
```
jorgeantonio21, line 368: This is great to keep track. I have a question, if a request is malformed, say because it is requesting a model that it is not supported by the node, or any other reason, this will count it too right ?

jorgeantonio21 (approved):
LGTM

atoma-service/src/handlers/chat_completions.rs
maschad: That's a good point, Axum handles issues at the HTTP request level, so it would throw before then, but I think if the model is unsupported this would be tracked here, I think we can refine this as we go on to make more granular metrics.

PR #396, opened 2025-02-07
feat: add country isocode validation to the atoma p2p service

jorgeantonio21:


atoma-bin/atoma_node.rs
```diff
@@ -212,6 +212,8 @@ async fn main() -> Result<()> {
         event = "p2p_node_spawn",
         "Spawning Atoma's p2p node service"
     );
+    validate_country_code(&config.p2p.country)?;
```
Cifko, line 215: you can use the validator crate to add this to the config directly, just slightly change the result of the function to ValidationError. We should add also the public url validtor directly to the config. That way you can just look at the config struct and see which validators do we have in place.

```
    #[validate(custom(function = "validate_country_code"))]
    pub country: String,
 ```

maschad (commented):
Agree with @Cifko 's suggested approach [here](https://github.com/atoma-network/atoma-node/pull/396/files#r1946532552)

Cifko (approved):

PR #397, opened 2025-02-08
fix: add p2p config fixes

maschad:
We were having issues listening on the config address as it was a TCP configered address.

I've also added the TCP behaviour as this will be useful in Kademlia for connecting to peers which may not be QUIC-enabled

### DIFF
--- a/atoma-p2p/src/service.rs
+++ b/atoma-p2p/src/service.rs
@@ -3,13 +3,14 @@ use crate::{
     timer::{usage_metrics_timer_task, NodeUsageMetrics},
     types::{AtomaP2pEvent, GossipMessage, UsageMetrics},
 };
+use config::ConfigError;
 use flume::Sender;
 use futures::StreamExt;
 use libp2p::{
     gossipsub::{self, ConfigBuilderError},
-    identify, kad, mdns,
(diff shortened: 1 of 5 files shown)


atoma-p2p/src/service.rs
```diff
@@ -244,42 +250,38 @@ impl AtomaP2pNode {
...
-                target = "atoma-p2p",
-                event = "listen_on_error",
-                listen_addr = config.listen_addr,
-                error = %e,
-                "Failed to listen on address"
-            );
-            AtomaP2pNodeError::SwarmListenOnError(e)
-        })?;
+        swarm
+            .listen_on(config.listen_addr.parse().unwrap())
```
jorgeantonio21, line 254: Maybe use `expect` here ? Also might be good to handle the error instead

jorgeantonio21 (approved):
LGTM

atoma-p2p/src/service.rs
maschad: Good call, I've updated in cc5d4d5 If that works feel free to merge

jorgeantonio21 (approved):
LGTM

PR #398, opened 2025-02-10
feat: add more detailed node metrics for p2p broadcasting

jorgeantonio21:


atoma-p2p/src/metrics.rs
```diff
@@ -0,0 +1,123 @@
...
+    let num_cpus = sys.cpus().len();
+    let networks = Networks::new_with_refreshed_list();
+    let mut network_rx = 0;
+    let mut network_tx = 0;
+    for (_interface, data) in &networks {
+        network_rx += data.received();
+        network_tx += data.transmitted();
+    }
+
+    sys.refresh_all();
```
maschad, line 101: Perhaps we should move this to line 86 before we collect the metrics to avoid staleness

atoma-p2p/src/metrics.rs
```diff
@@ -0,0 +1,123 @@
...
+    let device_count = nvml.device_count()?;
+    let mut gpus = Vec::new();
+    for i in 0..device_count {
+        let device = nvml.device_by_index(i)?;
+        let Utilization { gpu, memory } = device.utilization_rates()?;
+        let MemoryInfo { used, total, free } = device.memory_info()?;
+        let temperature = device.temperature(TemperatureSensor::Gpu)?;
+        let power_usage = device.power_usage()?;
+        gpus.push(GpuMetrics {
+            utilization: gpu,
```
maschad, line 76: Why do we use the same value for `utilizations` and `percentage_time_gpu_execution` ?

atoma-p2p/src/metrics.rs
```diff
@@ -0,0 +1,123 @@
...
+use tracing::instrument;
+
+/// Structure to store the usage metrics for the node
+///
+/// This data is collected from the system and the GPU
+/// to be sent across the p2p network, for efficient request routing.
+#[derive(Debug, Serialize, Deserialize, Default)]
+pub struct NodeMetrics {
+    // /// The uptime of the node
+    // pub uptime: u64,
```
maschad, line 18: Should this remain commented out?

atoma-p2p/src/service.rs
```diff
@@ -1284,26 +1291,37 @@ mod tests {
...
+                node_public_url,
+                node_small_id,
+                country,
+                timestamp,
+            },
+            node_metrics: NodeMetrics::default(),
+        };
+
+        let mut node_message_bytes = Vec::new();
+        ciborium::into_writer(&node_message, &mut node_message_bytes)
```
maschad, line 1305: It looks like we are doing a lot of repeated serialization here perhaps we should create a trait, something like

```rust
#[derive(Debug)]
struct SerializedMessage {
    bytes: Vec<u8>,
    hash: blake3::Hash,
}

impl NodeMessage {
    fn serialize(&self) -> Result<SerializedMessage, AtomaP2pNodeError> {
        let mut bytes = Vec::with_capacity(1024);
        ciborium::into_writer(self, &mut bytes).map_err(|e| {
            AtomaP2pNodeError::UsageMetricsSerializeError(e)
        })?;
        Ok(SerializedMessage {
            hash: blake3::hash(&bytes),
            bytes,
        })
    }
}

fn handle_new_usage_metrics_event(&mut self, usage_metrics: NodeMessage) -> Result<(), AtomaP2pNodeError> {
    let serialized = usage_metrics.serialize()?;
    let active_sui_address = self.keystore.addresses()[0];
    let signature = self.keystore.sign_hashed(&active_sui_address, serialized.hash.as_bytes())?;

    let signed_message = SignedNodeMessage {
        node_message: usage_metrics,
        signature: signature.as_ref().to_vec(),
    };

    let topic = gossipsub::IdentTopic::new(METRICS_GOSPUBSUB_TOPIC);
    self.swarm.behaviour_mut().gossipsub.publish(topic, serialized.bytes)?;
    Ok(())
}

async fn validate_signed_node_message(
    signed_node_message: &SignedNodeMessage,
    state_manager_sender: &Sender<StateManagerEvent>,
) -> Result<(), AtomaP2pNodeError> {
    validate_node_message_country_url_timestamp(&signed_node_message.node_message)?;
    
    let serialized = signed_node_message.node_message.serialize()?;
    verify_signature(&signed_node_message.signature, serialized.hash.as_bytes())?;
    verify_node_small_id_ownership(
        signed_node_message.node_message.node_metadata.node_small_id,
        &signed_node_message.signature,
        state_manager_sender,
    ).await
}
```

maschad (changes requested):
Looks great @jorgeantonio21 left some comments

atoma-p2p/src/metrics.rs
jorgeantonio21: Good observation, I will keep only the `percentage_time_gpu_execution`

atoma-p2p/src/metrics.rs
jorgeantonio21: Definitely agree, great catch !

atoma-p2p/src/service.rs
jorgeantonio21: Nice ! Great suggestion, we can definitely compute the byte size of the ciborium of the node message, which we could use for `with_capacity` on the vector allocation

jorgeantonio21: Addressed the comment

maschad (approved):

PR #425, opened 2025-02-19
build: remove unnecessary volumes from dev compose

maschad:
Related to https://github.com/atoma-network/atoma-proxy/pull/243

We don't need volumes for short-lived environments


jorgeantonio21 (approved):
LGTM

PR #426, opened 2025-02-19
build: add grafana root url for deployed instances

maschad:


docker-compose.yaml
```diff
@@ -115,6 +115,8 @@ services:
     volumes:
       - grafana_data:/var/lib/grafana
     env_file: .env
+    environment:
+      - GF_SERVER_ROOT_URL=http://${GRAFANA_DOMAIN:-149.7.4.141}:${GRAFANA_PORT:-30001}/
```
jorgeantonio21, line 119: Why is `149.7.4.141` hardcoded ?

maschad: It just defaults to this, but we can of course pass in the our own IP via the `GRAFANA_DOMAIN` env var

If we don't want this codeified at all I can remove it.

jorgeantonio21: Yeah, I would suggest to not have any explicit hardcoded IP on it

docker-compose.dev.yaml
```diff
@@ -53,7 +53,7 @@ services:
       - POSTGRES_USER=${POSTGRES_USER}
       - POSTGRES_PASSWORD=${POSTGRES_PASSWORD}
     ports:
-      - "${POSTGRES_PORT:-5432}:5432"
+      - "${POSTGRES_PORT}:${POSTGRES_INTERNAL_PORT}"
```
jorgeantonio21, line 56: I think this is very dangereous, the internal postgres port is 5432. I will remove it

maschad: I'm not sure how this can be dangerous, please expound. This compose file is intended to only be used for CI runs, and I don't see how mapping a port to another value is dangerous.

PR #430, opened 2025-02-20
build: add internal port for CI runs

maschad:
This is needed for https://github.com/atoma-network/atoma-testground


jorgeantonio21 (approved):
LGTM

PR #434, opened 2025-02-21
fix: update otel-collector fix + fix docker warning

maschad:


atoma-bin/telemetry.rs
```diff
@@ -30,8 +30,8 @@ const NODE_LOG_FILE: &str = "atoma-node.log";
 const DAEMON_LOG_FILE: &str = "atoma-daemon.log";
 
 // Default Grafana OTLP endpoint if not specified in environment
-
-const DEFAULT_OTLP_ENDPOINT: &str = "http://localhost:4317";
+// Override this to be the localhost for local development if not using Docker
+const DEFAULT_OTLP_ENDPOINT: &str = "http://otel-collector:4317";
```
jorgeantonio21, line 34: I suggest not having this hardcoded here. It makes more sense to add it to the `config.toml`

Dockerfile
```diff
@@ -39,7 +39,7 @@ RUN if [ "$ENABLE_TDX" = "true" ]; then \
     fi
 
 # Final stage
-FROM --platform=$TARGETPLATFORM ubuntu:24.04
```
jorgeantonio21, line 42: No more platform support ?

jorgeantonio21 (commented):
Looks good, left some comments though

Dockerfile
maschad: Setting platform to predefined `$TARGETPLATFORM ` in `FROM` is redundant as this is the default behaviour

atoma-bin/telemetry.rs
maschad: I actually disagree since we want to default to the docker instance and if one wants to override it then they can use an env var, but I feel this setup is more frictionless

jorgeantonio21: Makes sense

jorgeantonio21 (approved):
Lgtm

PR #446, opened 2025-02-25
feat: refactor atoma p2p to use Bytes instead of Vec<u8>

jorgeantonio21:


maschad (approved):
Great work here @jorgeantonio21

PR #477, opened 2025-03-12
feat: add cert chain to key rotation

jorgeantonio21:


atoma-confidential/src/service.rs
```diff
@@ -131,16 +130,18 @@ impl AtomaConfidentialCompute {
     ) -> Result<Self> {
         let key_manager = X25519KeyPairManager::new()?;
         let mut is_cc_supported = false;
-        for index in &config.device_indices {
-            is_cc_supported &= check_confidential_compute_status(u32::from(*index))?;
+        // NOTE: If the NVML library is not available, we return 0 devices
+        let num_devices = num_devices().unwrap_or(0);
+        for index in 0..num_devices {
+            is_cc_supported &= check_confidential_compute_status(index)?;
```
Cifko, line 136: this is actually always false, the initial value should be true (but probably only in cases when the num_devices is at least 1

atoma-state/src/state_manager.rs
```diff
@@ -987,23 +988,26 @@ impl AtomaState {
...
             ) 
-            VALUES ($1, $2, $3, $4, $5, $6)
+            VALUES ($1, $2, $3, $4, $5, $6, $7)
             ON CONFLICT (node_small_id, device_type)
             DO UPDATE SET
                 epoch = $1,
                 key_rotation_counter = $2,
                 public_key_bytes = $4,
-                remote_attestation_bytes = $5
+                remote_attestation_bytes = $5,
```
Cifko, line 1001: btw instead of $5 on conflict you can do stuff like : `remote_attestation_bytes = EXCLUDED.remote_attestation_bytes` 
you don't have to change it, I just like it better, because it's easier to see it's the correct value

Cifko (approved):
address the comments and the you can push

atoma-confidential/src/nvml_cc.rs
```diff
@@ -84,11 +88,85 @@ pub fn fetch_attestation_report(index: usize, nonce: [u8; 32]) -> Result<Vec<u8>
...
+/// certificate chain from the specified GPU device. The certificate chain is used
+/// to verify the authenticity of attestation reports generated by the device.
+///
+/// # Arguments
+///
+/// * `index` - The index of the device to fetch the certificate chain for
+///
+/// # Returns
+///
+/// * `Vec<u8>` - The raw certificate chain data for the given device
```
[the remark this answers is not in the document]
jorgeantonio21: Done

atoma-confidential/src/service.rs
jorgeantonio21: Oh nice catch !

atoma-state/src/state_manager.rs
jorgeantonio21: Will change

PR #483, opened 2025-03-13
feat: pass nvidia devices to atoma-node in docker

Cifko:
Pass nvidia devices to atoma-node in the docker


jorgeantonio21 (approved):
Perfect,

LGTM

PR #484, opened 2025-03-14
feat: address parsing of key rotation counter and nonce in `get_last_key_rotation_event`

jorgeantonio21:
### Description

This PR fixes an issue in extracting the key_rotation_counter and nonce fields from the object.fields JSON representation.

### Changes Made

1. Corrected the extraction logic to first attempt parsing values as u64 and then as str before converting to u64.
2. Fixed redundant .and_then() chaining that could cause incorrect parsing behavior.
3. Ensured both key_rotation_counter and nonce are correctly extracted before returning them.

### Issue Fixed

1. Previously, the extraction attempted .as_u64() before .as_str(), which would always fail for string-encoded numbers. 
2. This PR corrects the order to properly handle string representations.

### Testing

Verified correct parsing behavior with different JSON inputs.

### Impact

This fix ensures robust parsing of key rotation counters and nonces, preventing potential failures in cryptographic operations or state transitions.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2052,7 +2052,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "18e4fdb82bd54a12e42fb58a800dcae6b9e13982238ce2296dc3570b92148e1f"
 dependencies = [
  "data-encoding",
- "syn 2.0.100",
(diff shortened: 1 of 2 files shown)


Cifko (approved):
Add PR description as well

PR #512, opened 2025-03-26
feat: resolve deadlock in dashmap

jorgeantonio21:


atoma-service/src/handlers/mod.rs
```diff
@@ -351,20 +351,23 @@ pub fn handle_concurrent_requests_count_decrement(
...
-            endpoint = endpoint,
-            "Concurrent requests count is 0 for stack small id: {}, but we still need to update the stack num compute units",
-            stack_small_id
-        );
-    }
-    *concurrent_requests_count = concurrent_requests_count.saturating_sub(1);
-    if *concurrent_requests_count == 0 {
+    let concurrent_requests_count = {
+        let mut concurrent_requests_count = concurrent_requests_per_stack
+            .entry(stack_small_id)
```
maschad, line 356: Should we keep this entry locked until the new count is returned? Otherwise another thread could modify the value.

```rust
let concurrent_requests_count = {
    // Use entry_ref() to avoid cloning the key
    let mut entry = concurrent_requests_per_stack.entry(stack_small_id);
    let mut count = entry.or_insert(0);
    
    if *count == 0 {
        tracing::error!(
            target = "atoma-service",
            level = "error",
            endpoint = endpoint,
            "Concurrent requests count is 0 for stack small id: {}, but we still need to update the stack num compute units",
            stack_small_id
        );
    }
    
    *count = count.saturating_sub(1);
    *count // Return the new value while still holding the lock
}; // Lock is released here

// Now use the result
if concurrent_requests_count == 0 {
    // etc
}
```

atoma-service/src/handlers/mod.rs
```diff
@@ -351,20 +351,23 @@ pub fn handle_concurrent_requests_count_decrement(
...
+            .or_insert(0);
+        if *concurrent_requests_count == 0 {
```
maschad, line 358: If the `concurrent_requests_count` is unexpectedly zero should we return early with specific error handling? or should we ser to a minimum valid value? I think continuing execution when the count is unexpectedly 0 might lead to further issues.

maschad (changes requested):
Nice work, left some comments

atoma-service/src/server.rs
```diff
@@ -230,37 +231,43 @@ pub fn create_router(app_state: AppState) -> Router {
...
-            ServiceBuilder::new()
-                .layer(from_fn(signature_verification_middleware))
-                .layer(from_fn_with_state(
-                    app_state.clone(),
-                    verify_stack_permissions,
-                ))
-                .into_inner(),
+        .route(IMAGE_GENERATIONS_PATH, post(image_generations_handler));
+
+    let public_routes = Router::new()
```
maschad, line 241: I believe these changes should be apart of https://github.com/atoma-network/atoma-node/pull/511

jorgeantonio21: Yeah, let's merge that one first pls

maschad (approved):
Awesome work on this @jorgeantonio21 very good find

Cifko (approved):

Cifko (approved):

PR #523, opened 2025-03-28
feat: add instrument errors

jorgeantonio21:


Cifko (approved):

atoma-service/src/handlers/chat_completions.rs
```diff
@@ -194,7 +194,8 @@ pub struct ChatCompletionsOpenApi;
 #[instrument(
     level = "info",
     skip(state, payload),
```
maschad, line 196: This is not directly included in this PR but should these be `skip_all` as well?

jorgeantonio21: Yes, good catch !

PR #539, opened 2025-04-03
build: allow for non-nvidia runtime builds

maschad:
This will allow us to run without the `nvidia` runtime for our e2e tests as orchestrated in https://github.com/atoma-network/atoma-testground/pull/17

I've also updated the `README.md` to reflect the reality that we always run in confidential mode.


jorgeantonio21 (approved):
LGTM

PR #544, opened 2025-04-07
feat: replace kv cache usage with waiting queue times vllm metrics

jorgeantonio21:


maschad (approved):

atoma-service/src/handlers/mod.rs
```diff
@@ -593,13 +594,13 @@ mod vllm_metrics {
...
-        if min_kv_cache_usage > 0.9 {
+        if min_request_queue_time_seconds > MAX_ALLOWED_REQUEST_QUEUE_TIME_SECONDS {
             tracing::warn!(
                 target = "atoma-service",
                 level = "warn",
-                "Best available chat completions service URL for model: {model} has a GPU cache usage of {min_kv_cache_usage}%",
+                "Best available chat completions service URL for model: {model} has a request queue time of at least {min_request_queue_time_seconds} seconds",
                 model = model,
-                min_kv_cache_usage = min_kv_cache_usage
+                min_request_queue_time_seconds = min_request_queue_time_seconds
```
Cifko, line 603: you don't have to do `min_request_queue_time_seconds  = min_request_queue_time_seconds` , because it's the same name, the same goes for `model = model`, it's required only for things like `model_name = model`

atoma-service/src/handlers/mod.rs
```diff
@@ -549,28 +549,29 @@ mod vllm_metrics {
         chat_completions_service_urls: &[String],
         model: &str,
     ) -> Result<(String, StatusCode)> {
+        const MAX_ALLOWED_REQUEST_QUEUE_TIME_SECONDS: f64 = 2.0;
```
Cifko, line 552: why is max time 2 when the min request queue time is 4 (change on line 558)
so the min is always larger than max?

atoma-service/src/handlers/mod.rs
jorgeantonio21: I like to have it because it shows the variable name before the value in it

Cifko: I don't like it :-)
it's like when you are creating a struct, you don't do `{ value : value} ` you just do `{value}` because it's less clutter.

jorgeantonio21: Actually, I remove it, it doesn't compile. I am going to remove the vars altogether

Issue #553, opened 2025-04-09
Separate confidential and regular endpoints metrics


jorgeantonio21:
To comparate added latency of encryption/decryption.

PR #557, opened 2025-04-09
feat: add confidential and non-confidential latency metrics

maschad:
- Closes https://github.com/atoma-network/atoma-node/issues/553

### DIFF
--- a/atoma-service/src/handlers/chat_completions.rs
+++ b/atoma-service/src/handlers/chat_completions.rs
@@ -117,7 +117,7 @@ const UNKNOWN_MODEL: &str = "unknown";
 /// - `Usage`: Token usage statistics
 /// - `Choice`: A single completion choice
 /// - `ChatCompletionsResponse`: The complete response structure
-/// - `ChatCompletionRequest`: The request body for chat completion requests        
+/// - `ChatCompletionRequest`: The request body for chat completion requests
 /// - `ChatCompletionMessage`: A message in the conversation (system, user, assistant, or tool)
 /// - `ChatCompletionResponse`: The complete response structure
 /// - `ChatCompletionChoice`: A single completion choice
@@ -1368,7 +1368,7 @@ pub mod utils {
         let response_body = match handle_confidential_compute_encryption_response(
             state,
             response_body,
-            client_encryption_metadata,
+            client_encryption_metadata.clone(),
             endpoint.clone(),
         )
         .await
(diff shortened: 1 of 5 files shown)


atoma-service/src/streamer.rs
```diff
@@ -342,6 +342,24 @@ impl Streamer {
 
         self.is_final_chunk_handled = true;
 
+        CHAT_COMPLETIONS_STREAMING_LATENCY_METRICS.record(
```
jorgeantonio21, line 345: This only records the total duration of the decoding phase. Let's also add metrics for this on time to first token and inter token latency.

atoma-service/src/handlers/chat_completions.rs
```diff
@@ -1377,7 +1377,7 @@ pub mod utils {
                 // Stop the timer before returning the valid response
                 CHAT_COMPLETIONS_LATENCY_METRICS.record(
                     timer.elapsed().as_secs_f64(),
-                    &[KeyValue::new("model", model.to_owned())],
+                    &[KeyValue::new("model", model.to_owned()), KeyValue::new("privacy_level", if client_encryption_metadata.is_some() { "confidential" } else { "non-confidential" })],
```
jorgeantonio21, line 1380: This has a different logical flow than embeddings and image gen, why ?

jorgeantonio21 (changes requested):

atoma-service/src/handlers/chat_completions.rs
maschad: I noticed that as well @jorgeantonio21 but it's correct in my view. 

[lines 1368-1374](https://github.com/maschad/atoma-node/blob/mc/feat/add-latency-measurement-encryption/atoma-service/src/handlers/chat_completions.rs#L1368-L1374) do a `match` case dependent on whether `client_encryption_metadata` is present (in which case indicates the request is a _confidential_ one) 

If it's not present, then [the response body is immediately returned](https://github.com/maschad/atoma-node/blob/mc/feat/add-latency-measurement-encryption/atoma-service/src/handlers/mod.rs#L235)  (which means it was _non-confidential_ request)

This approach varies to the image gen and embeddings which just utilize completely different handlers for confidential and non-confidential endpoints. Perhaps we could refactor to unify the logical flows across our service requests in the future.

jorgeantonio21: Makes sense to me !

jorgeantonio21 (approved):

PR #568, opened 2025-04-20
feat: update sui crates

jorgeantonio21:


maschad (changes requested):
Let's add these [packages](https://github.com/atoma-network/atoma-node/actions/runs/14558456086/job/40838567866?pr=568#step:4:23) to the [deny.toml](https://github.com/atoma-network/atoma-node/blob/main/deny.toml) since they are dependencies of the Sui rust ecosystem

maschad (approved):

PR #578, opened 2025-04-24
feat: add uniqueness to claimed stacks

jorgeantonio21:


atoma-state/src/types.rs
```diff
@@ -296,4 +296,7 @@ pub struct UpdateStackNumComputeUnitsAndClaimFunds {
 
     /// Indicates whether the stack is locked for claim
     pub is_locked_for_claim: bool,
+
+    /// Indicates whether the stack was claimed already or not
+    pub was_claimed: bool,
```
Cifko, line 301: this should be `was_locked_for_claim`

PR #579, opened 2025-04-24
feat: lock the concurrent request counter

Cifko:
Lock the concurrent request counter, so it should never call the claim stack twice.

### DIFF
--- a/atoma-state/src/handlers.rs
+++ b/atoma-state/src/handlers.rs
@@ -1,3 +1,5 @@
+use std::sync::Arc;
+
 use atoma_p2p::types::AtomaP2pEvent;
 use atoma_sui::events::{
     AtomaEvent, ClaimedStackEvent, NewStackSettlementAttestationEvent,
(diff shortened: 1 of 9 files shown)


atoma-state/src/handlers.rs
```diff
@@ -809,28 +812,43 @@ pub(crate) async fn handle_update_stack_num_compute_units_and_claim_funds(
...
+                        total_compute_units,
+                        RATIO_FOR_CLAIM_STACK_THRESHOLD,
+                        count as i64,
+                    )
+                    .await?,
+                count,
+            )
+        }
+        Entry::Vacant(_entry) => {
+            return Err(AtomaStateManagerError::StackNotFound);
```
jorgeantonio21, line 849: Why ? This specific error doesn't make much sense in this context..

atoma-state/src/handlers.rs
```diff
@@ -809,28 +812,44 @@ pub(crate) async fn handle_update_stack_num_compute_units_and_claim_funds(
...
+        }
+    };
+    drop(entry);
```
jorgeantonio21, line 852: nit: I usually prefer doing this into a inner scope, instead of using drop. Can we refactor it ?

jorgeantonio21 (approved):

atoma-state/src/handlers.rs
Cifko: ok, changed

atoma-state/src/handlers.rs
Cifko: We are looking for a stack in the dashmap, btw this should never trigger

jorgeantonio21 (approved):

PR #584, opened 2025-04-24
fix: concurrent release lock too soon

Cifko:
Fix the concurrency issue where the lock is dropped before the DB is updated for concurrent requests.

### DIFF
--- a/atoma-state/src/handlers.rs
+++ b/atoma-state/src/handlers.rs
@@ -820,7 +820,8 @@ pub(crate) async fn handle_update_stack_num_compute_units_and_claim_funds(
         "Processing update stack num compute units and claim funds"
     );
     let entry = concurrent_requests.entry(stack_small_id);
-    let concurrent_requests = match &entry {
+    // Extract the count of concurrent requests for this stack_small_id
+    let request_count = match &entry {
         Entry::Occupied(entry) => *entry.get(),
         Entry::Vacant(_entry) => 0, // If it was zero it was deleted, so vacant is treated as zero, don't do insert, because there is no cleanup
     };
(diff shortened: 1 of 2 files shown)


atoma-state/src/handlers.rs
```diff
@@ -819,30 +819,28 @@ pub(crate) async fn handle_update_stack_num_compute_units_and_claim_funds(
...
+    } = state_manager
+        .state
+        .update_stack_num_compute_units(
+            stack_small_id,
+            estimated_total_compute_units,
+            total_compute_units,
+            RATIO_FOR_CLAIM_STACK_THRESHOLD,
+            concurrent_requests as i64,
+        )
+        .await?;
```
maschad, line 842: For the sake of debugging could we add

```rust
{
    Ok(result) => result,
    Err(err) => {
        // Log detailed information about the error context
        tracing::error!(
            "Failed to update stack compute units for stack_id={}: error={:?}, concurrent_requests={}",
            stack_small_id,
            err,
            request_count
        );
        drop(entry); // Drop the lock before propagating the error
        return Err(err);
    }
};
```

around the database operation to make debugging easier if something fails.

atoma-state/src/handlers.rs
```diff
@@ -819,30 +819,28 @@ pub(crate) async fn handle_update_stack_num_compute_units_and_claim_funds(
...
-                    estimated_total_compute_units,
-                    total_compute_units,
-                    RATIO_FOR_CLAIM_STACK_THRESHOLD,
-                    count as i64,
-                )
-                .await?,
-            count,
-        )
+    let entry = concurrent_requests.entry(stack_small_id);
+    let concurrent_requests = match &entry {
```
maschad, line 823:
```suggestion
// Extract the count of concurrent requests for this stack_small_id
let request_count = match &entry {
    Entry::Occupied(entry) => *entry.get(),
    Entry::Vacant(_entry) => 0, // If it was zero it was deleted, so vacant is treated as zero
};
```

maschad (changes requested):
Nice work, left some comments

atoma-state/src/handlers.rs
Cifko: Done

atoma-state/src/handlers.rs
Cifko: Done

jorgeantonio21 (approved):
Great work on this

PR #586, opened 2025-04-25
feat: add fiat support

Cifko:
Add fiat request for allowed sui addresses.

### DIFF
--- a/atoma-service/src/config.rs
+++ b/atoma-service/src/config.rs
@@ -56,6 +56,9 @@ pub struct AtomaServiceConfig {
 
     /// The interval for updating the metrics.
     pub metrics_update_interval: Option<u64>,
+
+    /// List of allowed sui addresses for fiat payments.
+    pub whitelist_sui_addresses_for_fiat: Vec<String>,
 }
 
 impl AtomaServiceConfig {
--- a/atoma-service/src/handlers/chat_completions.rs
+++ b/atoma-service/src/handlers/chat_completions.rs
@@ -5,7 +5,7 @@ use crate::{
             CHAT_COMPLETIONS_CONFIDENTIAL_NUM_REQUESTS, CHAT_COMPLETIONS_ESTIMATED_TOTAL_TOKENS,
             TOTAL_FAILED_CHAT_CONFIDENTIAL_REQUESTS, TOTAL_FAILED_CHAT_REQUESTS,
         },
-        sign_response_and_update_stack_hash, update_stack_num_compute_units,
+        sign_response_and_update_stack_hash, update_fiat_amount, update_stack_num_compute_units,
     },
     middleware::EncryptionMetadata,
     server::AppState,
@@ -209,6 +209,8 @@ pub async fn chat_completions_handler(
     let RequestMetadata {
         stack_small_id,
         estimated_total_compute_units,
+        price_per_one_million_compute_units,
+        user_address,
         num_input_tokens,
         payload_hash,
         client_encryption_metadata,
@@ -237,6 +239,8 @@ pub async fn chat_completions_handler(
         endpoint.clone(),
         payload_hash,
         stack_small_id,
+        price_per_one_million_compute_units,
+        user_address.clone(),
         is_stream,
         payload.clone(),
         num_input_tokens,
@@ -265,19 +269,30 @@ pub async fn chat_completions_handler(
             // will not be penalized for the request.
             //
             // NOTE: We also decrement the concurrent requests count, as we are done processing the request.
-            let concurrent_requests = handle_concurrent_requests_count_decrement(
-                &state.concurrent_requests_per_stack,
-                stack_small_id,
-                "chat-completions/chat_completions_handler",
-            );
-            update_stack_num_compute_units(
-                &state.state_manager_sender,
-                stack_small_id,
-                estimated_total_compute_units,
-                0,
-                &endpoint,
-                concurrent_requests,
-            )?;
+            if let Some(stack_small_id) = stack_small_id {
+                let concurrent_requests = handle_concurrent_requests_count_decrement(
+                    &state.concurrent_requests_per_stack,
+                    stack_small_id,
+                    "chat-completions/chat_completions_handler",
+                );
+                update_stack_num_compute_units(
+                    &state.state_manager_sender,
+                    stack_small_id,
+                    estimated_total_compute_units,
+                    0,
+                    &endpoint,
+                    concurrent_requests,
+                )?;
+            } else {
+                update_fiat_amount(
+                    &state.state_manager_sender,
+                    user_address,
+                    estimated_total_compute_units,
+                    0,
+                    price_per_one_million_compute_units,
+                    &endpoint,
+                )?;
+            }
             match e {
                 // We want to propagate the error if the inference service is unavailable
                 AtomaServiceError::ChatCompletionsServiceUnavailable { .. } => Err(e),
@@ -394,6 +409,8 @@ pub async fn confidential_chat_completions_handler(
         stack_small_id,
         num_input_tokens,
         estimated_total_compute_units,
+        user_address,
+        price_per_one_million_compute_units,
         payload_hash,
         client_encryption_metadata,
         ..
@@ -426,6 +443,8 @@ pub async fn confidential_chat_completions_handler(
         endpoint.clone(),
         payload_hash,
         stack_small_id,
+        price_per_one_million_compute_units,
+        user_address.clone(),
         is_stream,
         payload.clone(),
         num_input_tokens,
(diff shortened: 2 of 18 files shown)


atoma-service/src/config.rs
```diff
@@ -53,6 +53,9 @@ pub struct AtomaServiceConfig {
 
     /// The environment for the Atoma Service.
     pub environment: Option<String>,
+
+    /// List of allowed sui addresses for fiat payments.
+    pub fiat_allowed: Vec<String>,
```
maschad, line 58: can we make this a bit more descriptive? such as  `fiat_allow_listed_addresses` upon reading this I had thought this was to enable/disable fiat payments for everyone (not just select wallets)

atoma-service/src/handlers/embeddings.rs
```diff
@@ -137,19 +140,30 @@ pub async fn embeddings_handler(
...
+                update_stack_num_compute_units(
+                    &state.state_manager_sender,
+                    stack_small_id,
+                    estimated_total_compute_units,
+                    0,
+                    &endpoint,
+                    concurrent_requests,
+                )?;
+            } else {
+                update_fiat_amount(
```
maschad, line 158: This calculation _could_ potentially lead to integer overflow if `estimated_total_compute_units` and `price_per_one_million_compute_units` are large enough integers. The integer overflow in release mode will wrap around, which could lead to incorrect financial calculations. Let's use `u128` during the calculations

```rust
// Rearranged calculation to prevent potential integer overflow
    // First divide to reduce the magnitude, then multiply
    let fiat_amount = (estimated_total_compute_units as u128 / ONE_MILLION as u128) 
        * price_per_one_million_compute_units as u128;
    
    update_fiat_amount(
        &state.state_manager_sender,
        user_address,
        fiat_amount,
        0,
        &endpoint,
    )?;
```

atoma-service/src/handlers/image_generations.rs
```diff
@@ -136,19 +138,30 @@ pub async fn image_generations_handler(
...
+                update_stack_num_compute_units(
+                    &state.state_manager_sender,
+                    stack_small_id,
+                    estimated_total_compute_units,
+                    0,
+                    &endpoint,
+                    concurrent_requests,
+                )?;
+            } else {
+                update_fiat_amount(
```
maschad, line 156: likewise

atoma-service/src/handlers/image_generations.rs
```diff
@@ -258,19 +273,30 @@ pub async fn confidential_image_generations_handler(
...
```
maschad, line 291: likewise

atoma-service/src/handlers/mod.rs
```diff
@@ -78,25 +80,25 @@ async fn sign_response_and_update_stack_hash(
...
-        .send(AtomaAtomaStateManagerEvent::UpdateStackTotalHash {
-            stack_small_id,
-            total_hash: total_hash_bytes,
-        })
-        .map_err(|e| AtomaServiceError::InternalError {
-            message: format!("Error updating stack total hash: {}", e),
-            endpoint: endpoint.clone(),
-        })?;
-
+    if let Some(stack_small_id) = stack_small_id {
```
maschad, line 84: If the hash length doesn't match exactly 32 bytes, the program will panic with the message "Invalid BLAKE2b hash length", causing an unexpected termination.
I would say using `expect()` in production code is generally discouraged because it doesn't allow for graceful error handling, lets map the errors.

```rust
if let Some(stack_small_id) = stack_small_id {
    // Update the stack total hash
    let total_hash = blake2b_hash(&[payload_hash, response_hash].concat());
    
    // Handle potential incorrect hash length with proper error
    let total_hash_bytes: [u8; 32] = total_hash
        .as_slice()
        .try_into()
        .map_err(|_| AtomaServiceError::InternalError {
            message: "Invalid BLAKE2b hash length".to_string(),
            endpoint: endpoint.clone(),
        })?;
    
    state
        .state_manager_sender
        .send(AtomaAtomaStateManagerEvent::UpdateStackTotalHash {
            stack_small_id,
            total_hash: total_hash_bytes,
        })
        .map_err(|e| AtomaServiceError::InternalError {
            message: format!("Error updating stack total hash: {}", e),
            endpoint: endpoint.clone(),
        })?;
}
```

maschad (changes requested):
Great work on this @Cifko 🚀 

Left some comments

atoma-service/src/streamer.rs
```diff
@@ -311,48 +319,65 @@ impl Streamer {
...
-            &self.endpoint,
-        );
-        if let Err(e) = update_stack_num_compute_units(
+            // Update stack num tokens
+            let concurrent_requests = handle_concurrent_requests_count_decrement(
+                &self.concurrent_requests,
+                stack_small_id,
+                &self.endpoint,
+            );
+            if let Err(e) = update_stack_num_compute_units(
```
jorgeantonio21, line 353: Do we want to unlock possible locked payments if something fails ?

atoma-service/src/middleware.rs
```diff
@@ -443,119 +449,182 @@ pub async fn verify_stack_permissions(
...
-            return Err(AtomaServiceError::UnavailableStackError {
-                message: format!(
-                    "Stack with stack_small_id={stack_small_id} is unavailable, please buy a new stack or await it to be available again."
+            })?
+            .map_err(|err| AtomaServiceError::AuthError {
+                auth_error: format!(
+                    "Failed to get available stack with enough compute units, with error: {err}"
                 ),
                 endpoint: endpoint.clone(),
+            })?;
```
jorgeantonio21, line 496: All this logic could be encapsulated in a function,  and we should have one for fiat and stack payments.

atoma-service/src/handlers/mod.rs
jorgeantonio21: I believe there is literally 0 chance this happens, but we should not expect I agree.

atoma-service/src/handlers/chat_completions.rs
```diff
@@ -1062,7 +1108,7 @@ pub mod utils {
                 .map_err(|e| {
                     AtomaServiceError::InternalError {
                         message: format!(
-                            "Error sending encryption request, for request with payload hash: {:?}, and stack small id: {}, with error: {}",
+                            "Error sending encryption request, for request with payload hash: {:?}, and stack small id: {:?}, with error: {}",
```
jorgeantonio21, line 1111: why ?

atoma-service/src/handlers/embeddings.rs
jorgeantonio21: This makes sense, we should address it everywhere in the code.

atoma-service/src/config.rs
Cifko: renamed to `fiat_allowed_addresses`

atoma-service/src/handlers/embeddings.rs
Cifko: I will change this, but just for fun, if the price for 1M tokens is $1000, and we do the same as usdc, 6 decimal places for the USD balance. It will overflow it there are estimation for more than 9B tokens, that's how big the i64 is :-)

Cifko: also I don't want to do the division first, because that will reduce the price to zero, if the token usage is e.g. 900k (which is not) then the equation above is zero, but I agree to put 128bit number just to be sure, I hope clippy doesn't complain about converting u128 back to i64

atoma-state/src/state_manager.rs
```diff
@@ -1842,6 +1843,154 @@ impl AtomaState {
...
+    pub async fn update_fiat_amount(
+        &self,
+        user_address: String,
+        estimated_total_amount: i64,
+        total_amount: i64,
+    ) -> Result<()> {
+        sqlx::query(
+            "UPDATE fiat_balance 
+                  SET overcharged_unsettled_amount = fiat_balance.overcharged_unsettled_amount - $2,
+                      already_debited_amount = fiat_balance.already_debited_amount + $3,
```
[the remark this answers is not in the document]
Cifko: no, because we don't have a locking issue on fiat, the problem on stacks was you can run out of money on it and lock it. This will never happen to fiat.

atoma-state/src/migrations/20250430080652_fiat_balance.sql
```diff
@@ -0,0 +1,7 @@
+CREATE TABLE
+  IF NOT EXISTS fiat_balance (
+    user_address TEXT NOT NULL PRIMARY KEY,
+    already_debited_amount BIGINT NOT NULL DEFAULT 0,
+    overcharged_unsettled_amount BIGINT NOT NULL DEFAULT 0,
+    num_requests BIGINT NOT NULL DEFAULT 0
```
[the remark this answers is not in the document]
Cifko: maybe client addresses? because I suppose we will add our addresses for testing :-)

atoma-state/src/handlers.rs
```diff
@@ -767,6 +767,34 @@ pub(crate) async fn handle_state_manager_event(
                 .update_stack_total_hash(stack_small_id, total_hash)
                 .await?;
         }
+        AtomaAtomaStateManagerEvent::GetModelPricing {
```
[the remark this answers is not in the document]
Cifko: The node has it, but we receive only model name, so I have to grab the pricing the DB

atoma-service/src/streamer.rs
Cifko: in reality the locked payments don't have to be there, the locking was introduced so we don't overshoot the stack. But never happens to fiat. But I've added this to see if there is inference running and for how much. It's just an informative column.

atoma-service/src/handlers/chat_completions.rs
Cifko: the stack_id is optional, I agree I should change to either stack or user_address, because one is being filled.

atoma-service/src/handlers/mod.rs
Cifko: this was already here, but yeah, agreed, changed

atoma-service/src/middleware.rs
Cifko: done

jorgeantonio21 (approved):

PR #594, opened 2025-04-28
chore: suppress unavailable Chat completions and track frequency

maschad:
This is as a result of us receiving too many queries, `Chat completions service is unavailable: Chat completions service is unavailable` which means we have too much load. This should just be a warning


Cifko (commented):
I'm not sure if this will make it better. It will just prolong the time in proxy, instead of returning too many request, and open router rerouting the request to other provider.

jorgeantonio21 (changes requested):
This is not good, we should immediately return too many requests status code to the client.

maschad:
Closing in favour of https://github.com/atoma-network/atoma-node/pull/598

PR #595, opened 2025-04-29
feat: join metric queries for prometheus into one

Cifko:
Join prometheus queries.

### DIFF
--- a/atoma-service/src/handlers/mod.rs
+++ b/atoma-service/src/handlers/mod.rs
@@ -512,21 +512,19 @@ mod vllm_metrics {
                             .as_vector()
                             .ok_or_else(|| VllmMetricsError::NoMetricsFound(job.to_string()))
                             .and_then(|vector| {
-                                let x = vector
+                                vector
                                     .iter()
                                     .find(|instant| {
                                         instant.metric().get("job") == Some(&job.to_string())
                                     })
                                     .ok_or_else(|| {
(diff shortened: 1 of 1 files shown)


atoma-service/src/handlers/mod.rs
```diff
@@ -473,54 +472,65 @@ mod vllm_metrics {
...
+                            .and_then(|vector| {
+                                let x = vector
+                                    .iter()
+                                    .find(|instant| {
+                                        instant.metric().get("job") == Some(&job.to_string())
+                                    })
+                                    .ok_or_else(|| {
+                                        VllmMetricsError::NoMetricsFound(job.to_string())
+                                    });
+                                dbg!(&x);
```
jorgeantonio21, line 523: Remove

atoma-service/src/handlers/mod.rs
```diff
@@ -473,54 +472,65 @@ mod vllm_metrics {
...
-                    |value| {
-                        let sample = value.sample();
-                        let value = sample.value();
-                        Ok((chat_completions_service_url.to_string(), value))
-                    },
-                )
-            },
-        )
+        let response = client.query(&query).get().await;
+        jobs_with_url
```
jorgeantonio21, line 503: Do we need to change the way the config.toml is structured ?

Cifko: no, config is good

PR #598, opened 2025-04-29
fix: add prom metrics querying to background tasks

maschad:
Through our tests we found that  prometheus was failing to query the vllm for metrics, this is because for each request prometheus had to run a query to find the availability of the last 30 seconds. This is unreasonable because realistically this statistic would only update at less frequent intervals, and so we could just use the recently cached value.

So we run the metrics as a background task that updates this cache regularly. If no error is returned we proceed with the chat completion request.

### DIFF
--- a/atoma-service/src/handlers/mod.rs
+++ b/atoma-service/src/handlers/mod.rs
@@ -502,14 +502,22 @@ pub mod vllm_metrics {
     static METRICS_CACHE: LazyLock<MetricsCache> = LazyLock::new(|| MetricsCache::new());
 
     /// Start the background task to update metrics every 30 seconds
-    pub fn start_metrics_updater(chat_completions_service_urls: Vec<(String, String)>) {
+    pub fn start_metrics_updater(
+        chat_completions_service_urls: Vec<(String, String)>,
+        metrics_update_interval: Option<u64>,
+    ) {
         let chat_completions_service_urls = Arc::new(chat_completions_service_urls);
         tokio::spawn(async move {
-            let mut interval = time::interval(Duration::from_secs(30));
+            let metrics_interval = metrics_update_interval.unwrap_or(30);
+            let mut interval = time::interval(Duration::from_secs(metrics_interval));
             loop {
                 interval.tick().await;
                 let metrics = get_metrics(&HTTP_CLIENT, &chat_completions_service_urls).await;
-                METRICS_CACHE.update_metrics(metrics).await;
+                if metrics.iter().any(std::result::Result::is_ok) {
+                    METRICS_CACHE.update_metrics(metrics).await;
+                } else {
+                    tracing::warn!("Failed to retrieve any valid metrics, not updating cache");
+                }
             }
         });
     }
--- a/atoma-service/src/handlers/mod.rs
+++ b/atoma-service/src/handlers/mod.rs
@@ -616,9 +616,8 @@ pub mod vllm_metrics {
                 }
             };
             if request_queue_time_seconds.is_nan() {
-                continue;
-            }
-            if request_queue_time_seconds < min_request_queue_time_seconds {
+                min_request_queue_time_seconds = 0.0;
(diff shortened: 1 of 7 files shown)


atoma-service/src/handlers/mod.rs
```diff
@@ -602,7 +622,7 @@ mod vllm_metrics {
             model = model,
             best_url = best_url
         );
-        if min_request_queue_time_seconds >= MAX_ALLOWED_REQUEST_QUEUE_TIME_SECONDS {
+        if min_request_queue_time_seconds > MAX_ALLOWED_REQUEST_QUEUE_TIME_SECONDS {
```
jorgeantonio21, line 625: I don't think we should do this, since we are always updating for inferior values, so if all the vllm instances have higher waiting times for the queue, this will never be triggered

atoma-service/src/handlers/request_model.rs
```diff
@@ -14,8 +14,10 @@ pub struct ComputeUnitsEstimate {
 }
 
 /// A trait for parsing and handling AI model requests across different endpoints (chat, embeddings, images).
+///
 /// This trait provides a common interface for processing various types of AI model requests
 /// and estimating their computational costs.
+///
```
jorgeantonio21, line 20: Why ?

jorgeantonio21 (changes requested):

atoma-service/src/handlers/mod.rs
maschad: The `min_request_queue_time_seconds` would be set to the current vllm value along as it's not `nan` , what we found is it wasn't getting updated and thus remained at 2 and triggered this error, whilst that may be due the prom query error due to overload, I think it's still safer to wait till it passes the threshold.

Cifko: The problem was (and still is) that if the prometheus query fails the metrics will be just Vec of errors. So that the `min_request_queue_time_seconds` is never updated.
We should not update the metrics_cache when the query itself fails.

maschad: Good point @Cifko updated in 9586e44

jorgeantonio21: I don't think this will make it, as this logic will never be triggered.  We will need to change the logic of the current method.

atoma-service/src/handlers/mod.rs
```diff
@@ -463,29 +475,54 @@ mod vllm_metrics {
...
+    static METRICS_CACHE: LazyLock<MetricsCache> = LazyLock::new(|| MetricsCache::new());
+
+    /// Start the background task to update metrics every 30 seconds
+    pub fn start_metrics_updater(
+        chat_completions_service_urls: Vec<(String, String)>,
+        metrics_update_interval: Option<u64>,
+    ) {
+        let chat_completions_service_urls = Arc::new(chat_completions_service_urls);
+        tokio::spawn(async move {
+            let metrics_interval = metrics_update_interval.unwrap_or(30);
```
jorgeantonio21, line 511: Make the `30` a constant

atoma-service/src/handlers/mod.rs
Cifko: And how about the `min_request_queue_time_seconds` it will never be above the `MAX_ALLOWED_REQUEST_QUEUE_TIME_SECONDS` because we update only to smaller values than the initial.
Simply said, this will never trigger. So maybe set it to None and if we found a value smaller or the `min_request_queue_time_seconds.unwrap()` is greater then update it. That way we will get the actual waiting time in the logs.

maschad: @Cifko I don't it should be set to `None` but rather `Infinity` which guaranteed to be greater than any real number, so the first valid request queue time will always be less than it (given it's a number and not `NaN` 

6090fa5

Cifko (approved):

PR #599, opened 2025-04-30
chore: update too many requests code

maschad:


atoma-service/src/handlers/mod.rs
```diff
@@ -615,9 +615,14 @@ pub mod vllm_metrics {
                     continue;
                 }
             };
+
+            // Handle NaN case first
             if request_queue_time_seconds.is_nan() {
-                min_request_queue_time_seconds = 0.0;
-            } else if request_queue_time_seconds < min_request_queue_time_seconds {
+                continue;
```
Cifko, line 621: if we have 7 servers and if 6 of them are NaN (idle) and one is like 4 sec, the 4sec will be selected. That's bad.

maschad: This should have remained 0 seconds, my bad.

atoma-service/src/handlers/mod.rs
```diff
@@ -615,9 +615,14 @@ pub mod vllm_metrics {
                     continue;
                 }
             };
+
+            // Handle NaN case first
             if request_queue_time_seconds.is_nan() {
                 min_request_queue_time_seconds = 0.0;
```
Cifko, line 621: btw this should update the best_url as well, I missed that in the previous PR,

PR #609, opened 2025-05-07
fix: Remove unused `total_hash` column

maschad:
This was legacy code that was bloating our db volumes. This PR drops the unused total_hash field which is no longer needed since we are no longer relying on cross validation sampling consensus. 

- related https://github.com/atoma-network/atoma-proxy/pull/457


atoma-sui/src/client.rs
```diff
@@ -674,8 +674,6 @@ impl Client {
         stack_small_id: u64,
         node_badge_id: Option<ObjectID>,
         num_claimed_compute_units: u64,
-        committed_stack_proof: Vec<u8>,
```
jorgeantonio21, line 677: This will panic when dealing with the smart contract

jorgeantonio21 (changes requested):

jorgeantonio21 (approved):

PR #611, opened 2025-05-09
feat: handle completions endpoint

jorgeantonio21:


atoma-service/src/handlers/completions.rs
```diff
@@ -0,0 +1,1770 @@
...
+        headers,
+    )
+    .await
+    {
+        Ok(response) => {
+            CHAT_COMPLETIONS_ESTIMATED_TOTAL_TOKENS.add(
+                estimated_total_compute_units,
+                &[KeyValue::new("model", model.to_owned())],
+            );
+            TOTAL_COMPLETED_REQUESTS.add(1, &[KeyValue::new("model", model.to_owned())]);
```
Cifko, line 414: This should not be updated for streaming.

Cifko (approved):

atoma-service/src/handlers/completions.rs
jorgeantonio21: Why ?

PR #613, opened 2025-05-10
chore: attempt to parse error messages

maschad:
So that we can identify particular errors from vLLM instances.  Ideally the repetitve nature of this code could be refactored but it's a bit tricky given we consume the response later down in the control flow. So it would be a larger refactor.

### DIFF
--- a/atoma-service/src/handlers/chat_completions.rs
+++ b/atoma-service/src/handlers/chat_completions.rs
@@ -826,11 +826,36 @@ async fn handle_streaming_response(
         })?;
 
     if !response.status().is_success() {
-        let error = response
-            .status()
-            .canonical_reason()
-            .unwrap_or("Unknown error");
-        handle_status_code_error(response.status(), &endpoint, error)?;
+        let status = response.status();
+        let bytes = response
+            .bytes()
+            .await
+            .map_err(|e| AtomaServiceError::InternalError {
+                message: format!("Failed to read response body: {}", e),
+                endpoint: endpoint.to_string(),
+            })?;
+        // Try to parse the error message from the response body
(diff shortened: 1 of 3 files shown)


atoma-service/src/handlers/completions.rs
```diff
@@ -1182,26 +1206,56 @@ pub mod utils {
...
+                },
+                |json| {
+                    json.get("error")
+                        .or_else(|| json.get("message"))
+                        .and_then(|v| v.as_str())
+                        .unwrap_or_else(|| status.canonical_reason().unwrap_or("Unknown error"))
+                        .to_string()
+                },
+            );
+            handle_status_code_error(status, endpoint, &error_message)?;
```
jorgeantonio21, line 1233: We should add some info on the error message informing it comes from vllm, or any other llm inference service, so that we can identify it more easily.

atoma-service/src/handlers/chat_completions.rs
```diff
@@ -826,11 +826,36 @@ async fn handle_streaming_response(
...
+        let status = response.status();
+        let bytes = response
+            .bytes()
+            .await
+            .map_err(|e| AtomaServiceError::InternalError {
+                message: format!("Failed to read response body: {}", e),
+                endpoint: endpoint.to_string(),
+            })?;
+        // Try to parse the error message from the response body
+        let error_message = serde_json::from_slice::<Value>(&bytes).map_or_else(
```
jorgeantonio21, line 838: Is this really necessary ? Isn't there a way to plug the vllm/sglang logs into loki directly ?

jorgeantonio21 (commented):
Looks good, but address the following comments.

Was this tested ?

jorgeantonio21 (approved):

PR #622, opened 2025-05-21
feat: get metrics directly

Cifko:
Metrics are now not periodic, but on each request, directly from the inference service. We don't use TTFT or queue time anymore. Just the number of running and number of queued requests.

### DIFF
--- a/atoma-service/src/handlers/mod.rs
+++ b/atoma-service/src/handlers/mod.rs
@@ -551,8 +551,6 @@ pub mod inference_service_metrics {
     /// The default interval for updating the metrics
     const DEFAULT_METRICS_UPDATE_INTERVAL_MILLIS: u64 = 1_000;
 
-    /// The timeout for the Prometheus metrics queries
-    const METRICS_TIMEOUT: std::time::Duration = std::time::Duration::from_secs(2);
     /// Chat completions metrics
     #[derive(Debug, Clone)]
     struct ChatCompletionsMetrics {
@@ -659,12 +657,6 @@ pub mod inference_service_metrics {
                         &sglang_chat_completions_service_urls,
                     )
                     .await;
-                    info!(
-                        target = "atoma-service",
-                        module = "inference_service_metrics",
-                        level = "info",
-                        "Received SgLang metrics response for {sglang_chat_completions_service_urls:?}, {sglang_metrics:?}"
-                    );
                     if sglang_metrics.iter().any(std::result::Result::is_ok) {
                         SGLANG_METRICS_CACHE.update_metrics(sglang_metrics).await;
                     } else {
--- a/atoma-service/src/handlers/mod.rs
+++ b/atoma-service/src/handlers/mod.rs
@@ -50,6 +50,12 @@ pub const PROMPT_TOKENS_KEY: &str = "prompt_tokens";
 /// Key for the completion tokens in the usage in the response body
 pub const COMPLETION_TOKENS_KEY: &str = "completion_tokens";
 
+#[derive(Debug, Clone)]
+pub enum InferenceService {
+    Vllm,
+    SgLang,
+}
+
 /// Updates response signature and stack hash state
 ///
 /// # Arguments
@@ -517,8 +523,11 @@ pub fn handle_status_code_error(
 }
 
 pub mod inference_service_metrics {
+    use futures::future::join_all;
     use opentelemetry::KeyValue;
-    use prometheus_http_query::response::PromqlResult;
+    use prometheus_parse::Scrape;
+    use prometheus_parse::Value;
+    use rand::seq::SliceRandom;
     use rand::Rng;
     use std::sync::Arc;
     use std::sync::LazyLock;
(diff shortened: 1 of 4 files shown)


atoma-service/src/handlers/mod.rs
```diff
@@ -1055,46 +742,14 @@ pub mod inference_service_metrics {
         );
 
         // Get cached metrics
```
maschad, line 744: Remove this comment since we are no longer using a cache

maschad (changes requested):
Great work on this @Cifko 🚀  

I left some comments

maschad (approved):
Awesome :)

atoma-service/src/handlers/mod.rs
```diff
@@ -517,472 +523,150 @@ pub fn handle_status_code_error(
...
-                                model: model.to_string(),
-                                chat_completions_service_url: url.to_string(),
-                                waiting_queue_time: qt,
-                                time_to_first_token_seconds: ttf,
-                                num_running_requests: nrr,
-                                num_queue_requests: nqr,
+            .map(|(chat_completions_service_url, job)| {
+                let semaphore = METRICS_SEMAPHORE.clone();
+                async move {
+                    let _permit = semaphore.acquire().await.unwrap();
```
jorgeantonio21, line 605: Don't unwrap pls

atoma-service/src/handlers/mod.rs
```diff
@@ -1020,9 +704,7 @@ pub mod inference_service_metrics {
         chat_completions_service_urls: &[(String, String)],
         model: &str,
     ) -> Result<(String, StatusCode)> {
-        const MAX_ALLOWED_NUM_QUEUE_REQUESTS: f64 = 12.0; // Default to 12 requests
-        const MAX_ALLOWED_TIME_TO_FIRST_TOKEN_SECONDS: f64 = 6.0; // Default to 4 seconds
-        const MAX_ALLOWED_WAITING_TIME_SECONDS: f64 = 4.0; // Default to 4 seconds
+        const MAX_ALLOWED_NUM_QUEUED_REQUESTS: f64 = 12.0; // Default to 12 requests
```
jorgeantonio21, line 707: Set it to 1.0

atoma-service/src/handlers/mod.rs
```diff
@@ -1045,56 +727,21 @@ pub mod inference_service_metrics {
...
-                    .collect();
-            get_metrics_sglang(
-                &HTTP_CLIENT,
-                &sglang_chat_completions_service_urls_with_model,
-            )
-            .await
-        };
+        let vllm_metrics =
+            get_metrics(InferenceService::Vllm, &vllm_chat_completions_service_urls).await;
+        let sglang_metrics = get_metrics(
```
jorgeantonio21, line 739: This doesn't make sense. Either you extract vLLM or SgLang, or both if it is really the case both services are being used, but you should check

atoma-service/src/handlers/mod.rs
```diff
@@ -517,472 +523,157 @@ pub fn handle_status_code_error(
...
+                        chat_completions_service_url,
+                        num_queued_requests,
+                        num_running_requests,
                     })
-                })
-            })
-            .collect()
+                }
+            });
+        join_all(tasks).await
```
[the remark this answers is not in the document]
jorgeantonio21: From what I see, this hasn't been addressed (unresolved)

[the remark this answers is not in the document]
maschad: @jorgeantonio21 I addressed this with @Cifko on Slack. But I believe the current approach in c802aca53f7ea796469d08ce617adeaa53940d6f is appropriate. 


Previously, because the concurrent requests lock was local to this function, we could have a scenario where for each AI service request will have that many concurrent requests e.g if it was 10, with 100 service requests we would have a theoretical max of 1000 requests.

Now that a global semaphore is used then the lock and subsequently concurrent requests are spread across the AI service requests.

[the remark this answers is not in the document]
jorgeantonio21: I was not referring to any lock. We don't even need a semaphore. At most you have 8 services to query metrics from, which is a very tiny task.

atoma-service/src/handlers/mod.rs
Cifko: I disagree, this will just polute code if unnecessary check.

jorgeantonio21: I disagree it is poluting the code. I would rather not have unnecessary checks and unnecessary logs that just makes life more difficult for who is operating the node logic.

Cifko: so you want to add check if the vector e.g. `vllm_chat_completions_service_urls` is empty, and not call `get_metrics` if it's empty? but calling `get_metrics` with an empty array just does something like this `vec![].iter().collect()` so just returns empty vec

atoma-service/src/handlers/mod.rs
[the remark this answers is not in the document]
jorgeantonio21: Hum I see, I had the impression the semaphore was being used within the `get_best_available_urls` and therefore could not be used across multiple requests. In that case, we should handle this scenario

PR #623, opened 2025-05-21
feat: increase max body size to 1gb

jorgeantonio21:


maschad (changes requested):
This seems fine to be, but you will need to adjust [this test](https://github.com/atoma-network/atoma-node/blob/main/atoma-service/src/tests.rs#L1006)

Also should this be based off `develop` as opposed to `main`?

maschad (approved):

PR #631, opened 2025-05-26
feat: add usage per model

Cifko:
Add usage per model for user.

### DIFF
--- a/atoma-state/src/handlers.rs
+++ b/atoma-state/src/handlers.rs
@@ -782,21 +782,37 @@ pub(crate) async fn handle_state_manager_event(
         }
         AtomaAtomaStateManagerEvent::UpdateFiatAmount {
             user_address,
+            model_name,
             estimated_input_amount,
             input_amount,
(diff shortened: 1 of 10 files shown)


atoma-state/src/handlers.rs
```diff
@@ -782,21 +782,37 @@ pub(crate) async fn handle_state_manager_event(
...
                 .update_fiat_amount(
-                    user_address,
+                    user_address.clone(),
                     estimated_input_amount,
                     input_amount,
                     estimated_output_amount,
                     output_amount,
                 )
                 .await?;
+            if output_amount > 0 {
```
jorgeantonio21, line 803: Even the prefill phase generates at least one new token. When can it `0`?

jorgeantonio21 (approved):

atoma-state/src/handlers.rs
Cifko: when something went wrong, there is a plenty of usage where we send 0

PR #633, opened 2025-05-27
fix: completions estimations not triggered

Cifko:
The token estimation on `Completion` was not triggered because it was marked as `NonInference`


jorgeantonio21 (approved):

maschad (approved):
I think we can directly merge this into `main` because it's a hotfix @Cifko @jorgeantonio21

PR #644, opened 2025-05-28
feat: shuffle jobs for homogeneous load balancing

jorgeantonio21:


maschad (changes requested):
My preference would be for https://github.com/atoma-network/atoma-node/pull/643 as I do believe we should filter out the URLs that are already at max capacity.

maschad (approved):

PR #651, opened 2025-06-02
feat: add mem usage

Cifko:
Add memory usage cap at 0.9


atoma-service/src/handlers/mod.rs
```diff
@@ -33,6 +33,12 @@ const CIPHERTEXT_KEY: &str = "ciphertext";
 /// The default max tokens for a chat completion request
 const DEFAULT_MAX_TOKENS: u64 = 8_192;
 
+/// The ceiling for memory usage, above which the service will not accept new requests
+const MEMORY_USAGE_CEILING: f64 = 0.9;
```
jorgeantonio21, line 37: I reckon these values should be set in the config.toml, as we will most likely need to tweak them

PR #665, opened 2025-06-05
fix: how check for too many requests is handled

Cifko:
Fix how the too many requests release is handled.
It should not remove the model from the `too_many_requests` just because the timer passed, other check has to pass as well.

### DIFF
--- a/atoma-service/src/middleware.rs
+++ b/atoma-service/src/middleware.rs
@@ -1674,7 +1674,6 @@ pub mod utils {
                         endpoint: endpoint.to_string(),
                     });
                 }
-                occupied_entry.remove();
             }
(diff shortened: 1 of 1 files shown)


atoma-service/src/middleware.rs
```diff
@@ -1674,14 +1674,14 @@ pub mod utils {
                         endpoint: endpoint.to_string(),
                     });
                 }
-                occupied_entry.remove();
```
jorgeantonio21, line 1677: Why ? The ellapsed time has passed, so you should simply remove it.

Cifko: if you remove it here, then how do I know if you have to check for lower threshold? you don't so you applied to all, which is bad, the flag should be turned off when both time pass and lower threshold is hit

PR #668, opened 2025-06-09
feat: requests batcher

Cifko:
This pull request introduces a new batching mechanism for inference requests to improve performance and scalability. The changes include adding a `RequestsBatcher` component, modifying the `AppState` to include the batching sender, and updating request handling logic in the `chat_completions` and `completions` handlers to use the batching mechanism instead of direct HTTP calls.

### DIFF
--- a/atoma-service/src/handlers/completions.rs
+++ b/atoma-service/src/handlers/completions.rs
@@ -7,6 +7,7 @@ use crate::{
             TOTAL_FAILED_CHAT_REQUESTS, TOTAL_LOCKED_REQUESTS, TOTAL_TOO_EARLY_REQUESTS,
             TOTAL_TOO_MANY_REQUESTS, TOTAL_UNAUTHORIZED_REQUESTS,
         },
+        request_batcher::InferenceRequest,
         sign_response_and_update_stack_hash, update_fiat_amount, update_stack_num_compute_units,
     },
     middleware::EncryptionMetadata,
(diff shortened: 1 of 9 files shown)


atoma-service/src/handlers/completions.rs
```diff
@@ -1322,16 +1348,45 @@ pub mod utils {
...
-            .post(format!("{}{}", completions_service_url, COMPLETIONS_PATH))
-            .json(&payload)
-            .send()
-            .await;
+        let (result_sender, result_receiver) = tokio::sync::oneshot::channel();
+        state.request_batcher_sender.send(InferenceRequest {
+            url: format!("{}{}", completions_service_url, COMPLETIONS_PATH),
+            payload: payload.clone(),
+            result_sender
+        }).map_err(|e| {
```
jorgeantonio21, line 1356: I suggest we encapsulate this into a handle error method because this is a lot of deduplication of verbose code

jorgeantonio21 (changes requested):
The logic looks good to me, but there are so many repetitive `map_err`, I would either have an explicit method to handle this (you can implement it for `Result<T, AtomaServiceError>`, e.g.) or a Rust macro

atoma-service/src/handlers/completions.rs
jorgeantonio21: I am fine with handling the `map_err` deduplication in another PR

